PR #11, opened 2023-04-06
Small set with 2 associated domains

dmarti:
This is a test of First-Party Sets, including one manager domain and two co-branded content sites.

### DIFF
--- a/first_party_sets.JSON
+++ b/first_party_sets.JSON
@@ -16,6 +16,17 @@
         "https://recommendation.technology": "Recommendation offerings site",
         "https://digitalcustomerprofiling.com": "Customer profiling service"
       }
+    },
+    {
+      "primary": "https://cafemedia.com",
+      "associatedSites": [
+        "https://cardsayings.net",
+        "https://nourishingpursuits.com"
+      ],
+      "rationaleBySite": {
+        "https://cardsayings.net": "Co-branded content site",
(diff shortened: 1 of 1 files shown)


sjledoux:
Hello, it appears that your submission is failing the PR-Actions due to "https://cardsayings.net", and "https://nourishingpursuits.com" not having accessible .well-known pages. Currently, "https://cafemedia.com" is failing this check as well, likely due to a security feature on your site that blocks https requests made using urllib. That issue should be fixed from a push we just made to main on our end, so you will need to sync your PR with the base branch before that check passes.

dmarti:
@sjledoux Thank you, I have checked the member.json files for both member domains and they both seem to be up.

https://cardsayings.net/.well-known/first-party-set/member.json

https://nourishingpursuits.com/.well-known/first-party-set/member.json

sjledoux:
@dmarti Thank you for clarifying that. The issue would then be because we are expecting the member sites' JSON files to be served at /.well-known/first-party-set, **not** at /.well-known/first-party-set/member.json, causing those checks to fail.

dmarti:
Is a specific MIME type required for the JSON files on member sites? Is a redirect allowed?

sjledoux:
@dmarti I don't believe anything in the code requires a particular MIME type, so any should be fine. 
Redirects are allowed.

PR #23, opened 2023-05-02
added entry for AxelSpringerSE

renebaudisch:


sjledoux:
Hi @renebaudisch, thank you for your submission. In addition to the CLA check, it appears you're failing a number of the FPS submission checks, which you can view by clicking on the "Details" link on the right side of the tab where it says "FPS submission checks / PR-Actions (pull_request)".  

To summarize, it appears that a number of sites in your set are lacking the required .well-known/first-party-set.json file. Additionally, your service domain asadcn.com does not meet our requirements for service domains in regards to robots.txt and ads.txt.

sjledoux:
> What does that mean?
The service site https://www.asadcdn.com/ has an ads.txt file, this violates the policies for service sites

We are expecting that for each service domain, there is no ads.txt page. The error your PR received for that check suggests that when we made a request to https://asadcn.com/ads.txt we received a response code of 200, which constitutes a failure of the check.

> Where can I access those policies?

That policy is listed under the [Subset Level Technical Validation section of the submission guide.](https://github.com/GoogleChrome/first-party-sets/blob/main/FPS-Submission_Guidelines.md#subset-level-technical-validation)

> And I also don't understand what's wrong with our robots.txt, the header is set, and I cannot adjust them being case-sensitive

Thank you for pointing this out - our check here should not be case sensitive. I will create an issue in the repo with reference to this and fix it shortly.

renebaudisch:
Hi there, I did some changes and also see our next to does fixing the contents of the json files, but regarding the robots.txt error, I think there is a misspelling, shouldn't it search for `noindex` without `-`? Sure, I can add this value on my side also but won't the standard you will be faced give you `noindex` as well?

sjledoux:
Hi Renee, it looks like that's a misspelling in the error message; the code itself was looking at `noindex` without the `-`, which you can [see here](https://github.com/GoogleChrome/first-party-sets/blob/863f4cb010d36de9896dbbffb50f0040a79fe6fb/FpsCheck.py#L494). The reason why the response is showing up is because that version of the code was looking for **just** the `noindex` tag, while your site has `noindex, nofollow`. Your tags **should** be acceptable, so we are [fixing that](https://github.com/GoogleChrome/first-party-sets/blob/fdda2b5506856dbe57ac606fd7ecc54a10d8c6a0/FpsCheck.py#LL495C87-L495C87) with PR 28, as well as some of the wording with the [error message](https://github.com/GoogleChrome/first-party-sets/blob/fdda2b5506856dbe57ac606fd7ecc54a10d8c6a0/FpsCheck.py#L498) and the policy description in the submission guide. 

The fix will be merged shortly, so you should be passing the robots check soon. Hope this clears everything up in that regard. Apologies for the previous issues with that check.

renebaudisch:
No problem, we hopefully also updated our jsons till monday. 
Thanks for your fast responses and caring so much!

sjledoux:
Glad to have helped! 

By the way, for the changes you made on commits `45e82d7` and `a8b1d7f`, I noticed these look to be trivial. Was there any reason for these changes? Were you just making minor changes to trigger the workflow again?

renebaudisch:
We still need to update the json files, but the txt files should already be fixed. I purged the caches on the server to hopefully get rid of these warnings as they should be done....

sjledoux:
I believe you will need to rebase your branch to pass the robots check at least. When I run the current version of the code locally with your list, it is only failing the well-known checks.

ratchanonsuttawas:
renebaudisch:AxelSpringerSE_bild.de_init_submit

PR #24, opened 2023-05-02
Change workflow to run on PR request target

sjledoux:
Currently, the workflow is set to "on: pull_request" which does not allow the workflow to have write permission to PRs made from forked repos. This issue is preventing comments from appearing on those PRs. Therefore, the workflow should be set to "on: pull_request_target" which will allow the action to have the needed write permissions, and other changes will be made to ensure that other aspects of the workflow remain safe during this process.

### DIFF
--- a/.github/workflows/fps-submissions-checks.yml
+++ b/.github/workflows/fps-submissions-checks.yml
@@ -12,22 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 name: FPS submission checks
-on: [pull_request]
+on: pull_request_target
 jobs:
   PR-Actions:
     runs-on: ubuntu-latest
     outputs:
       output1: ${{ steps.read_results.outputs.contents }}
     steps:
-      - name: Gather files changed
+      - name: Get the PR version of the files
         uses: actions/checkout@v2
         with:
-          fileOutput: "json"
+          ref: "refs/pull/${{ github.event.number }}/merge"
+          path: "pull-request"
+      - name: Get the version at main
+        uses: actions/checkout@v2
+        with:
+          path: "main"
       - name: Get necessary libraries
         run: pip install publicsuffix2
       - name: Content check
         id: check
-        run: python3 check_sites.py > results.txt
+        run: python3 main/check_sites.py -i pull-request/first_party_sets.JSON > results.txt
       - name: Read the result
         id: read_results
         uses: andstor/file-reader-action@v1
@@ -46,8 +51,9 @@ jobs:
       - name: Write Comment on PR
         uses: mshick/add-pr-comment@v2
         with:   
-          message-path: 
-            "message.txt" 
+          message-path: "message.txt" 
+          allow-repeats: true
+          refresh-message-position: true
       - name: Fail or Succeed
         if: steps.read_results.outputs.contents != 'success'
         uses: actions/github-script@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/fps-submissions-checks.yml
```diff
@@ -46,8 +51,9 @@ jobs:
       - name: Write Comment on PR
         uses: mshick/add-pr-comment@v2
         with:   
-          message-path: 
-            "message.txt" 
+          message-path: "message.txt" 
+          allow-repeats: true
+          refresh-message-position: true
```
cfredric, line 56: It looks like `refresh-message-position` only has an effect if `allow-repeats` is false: https://github.com/mshick/add-pr-comment/blob/387ece43e304d694bf3f6f9bdc6188b1d7f1d90f/src/main.ts#L53-L79

So I would just use the `refresh-message-position` option, and remove the `allow-repeats`.

check_sites.py
```diff
@@ -13,19 +13,35 @@
...
+        try:
+            fps_sites = json.load(f)
+        except Exception as inst:
+        # If the file cannot be loaded, we will not run any other checks
+            print(inst)
+            exit()      
 
     # Load the etlds from the public suffix list
-    etlds = PublicSuffixList(psl_file = 'effective_tld_names.dat')
+    etlds = PublicSuffixList(psl_file = inputPrefix+'effective_tld_names.dat')
```
cfredric, line 41: Please use `os.path.join`, since that doesn't assume that the path separator should be `/`. Same comment in the other places that use `inputPrefix`.

check_sites.py
```diff
@@ -13,19 +13,35 @@
...
+    for opt, arg in opts:
+        if opt == '-i':
+            inputFile = arg
+            inputPrefix='main/'
+
     # Open the canonical sites, and load the json
-    with open('first_party_sets.JSON') as f:
-        fps_sites = json.load(f)
+    with open(inputFile) as f:
+        try:
```
cfredric, line 33: Since we're adding the exception handling here, can we also add exception handling to the code that parses the `.well-known/first-party-set.json` files?

check_sites.py
```diff
@@ -13,19 +13,35 @@
...
 
 def main():
+    args = sys.argv[1:]
+    inputFile = 'first_party_sets.JSON'
+    inputPrefix = ''
+    opts, _ = getopt.getopt(args, "i:")
+    for opt, arg in opts:
+        if opt == '-i':
+            inputFile = arg
+            inputPrefix='main/'
```
cfredric, line 29: Can we supply this as another argument (e.g. `data_directory`), rather than hardcoding it?

Alternatively, you could derive it from the path of this "binary", using `sys.executable` and `os.path.split`.

sjledoux: Applied the suggestion to supply this as argument "--data_directory"

check_sites.py
sjledoux: Applied this suggestion

.github/workflows/fps-submissions-checks.yml
sjledoux: Done

FpsCheck.py
```diff
@@ -329,8 +329,8 @@ def check_list_sites(self, primary, site_list):
                 if 'primary' not in json_schema.keys():
                     self.error_list.append(
                         "The listed associated site site did not have primary"
```
cfredric, line 331: FYI, Python supports multiline strings, which you could use instead of manually splitting/concatenating this string to avoid formatting issues. Same comment elsewhere.

cfredric (approved):
LGTM

PR #28, opened 2023-05-04
policy and code changes for robots check

sjledoux:
After investigating the [issue relating to case-insensitivity](https://github.com/GoogleChrome/first-party-sets/issues/25), it was determined that this was not the source of a problem with the robots check as our code uses python's urllib library for requests, which implements a case-insensitive dictionary for request headers. 

However, it was determined that there were some issues with the current implementation of the robots check. In particular, we realized that the non-existence of a robots.txt does not prevent web-crawlers from indexing a page. Additionally, the exiting method of checking the 'X-Robots-Tag' would exclude values of the tag that ought to be valid. This PR makes corresponding changes in the code and the submission guide to fix these issues.

### DIFF
--- a/FPS-Submission_Guidelines.md
+++ b/FPS-Submission_Guidelines.md
@@ -221,9 +221,10 @@ The /.well-known/first-party-set.json file for set members must follow the schem
 ## Subset-level technical validation ##
 
 Additionally, more granular technical checks will also run on GitHub for service domains and ccTLD variants in the submissions. 
+
 Service Domains must satisfy the following conditions:
 	<ul>
-		<li>Must not have robots.txt, or have a robots.txt with a ["no index"](https://developers.google.com/search/docs/advanced/crawling/block-indexing) header. </li>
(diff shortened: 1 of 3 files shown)


erikb-stripe:
I opened https://github.com/GoogleChrome/related-website-sets/issues/246 today. This change breaks for service domains that redirect to another domain that itself doesn't have a X-Robots-Tag.

Meinne1:
> After investigating the [issue relating to case-insensitivity](https://github.com/GoogleChrome/first-party-sets/issues/25), it was determined that this was not the source of a problem with the robots check as our code uses python's urllib library for requests, which implements a case-insensitive dictionary for request headers. 
> 
> However, it was determined that there were some issues with the current implementation of the robots check. In particular, we realized that the non-existence of a robots.txt does not prevent web-crawlers from indexing a page. Additionally, the exiting method of checking the 'X-Robots-Tag' would exclude values of the tag that ought to be valid. This PR makes corresponding changes in the code and the submission guide to fix these issues.

PR #32, opened 2023-05-24
Run checks on changes only

sjledoux:
Currently, the checks in this repo are run on all First-Party Set entries when the workflow is triggered. This creates cases where users who are trying to enter their First-Party Set to the list will have the workflow fail due to mistakes/lack of maintenance for an older set that they do not own. With this change, when the workflow is called with "with_diff=true", the checks will only be run on sets that have been added to the list or modified.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -93,7 +93,7 @@ class FpsCheck:
         self.error_list += load_sets_errors
         return check_sets
 
-    def has_all_rationales(self, check_sets):
+    def has_all_rationales(self):
         """Checks for the presence of all rationaleBySite elements in schema
 
         Reads the associated sites and service sites from all FpsSets, and 
@@ -103,13 +103,13 @@ class FpsCheck:
         error_list
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            None
         Returns:
             None
         """
         for fpset in self.fps_sites['sets']:
-            sites = check_sets[fpset.get("primary")].associated_sites
-            service_sites = check_sets[fpset.get("primary")].service_sites
+            sites = fpset.get("associatedSites")
+            service_sites = fpset.get("serviceSites")
             if sites:
                 sites = sites + service_sites if service_sites else sites
             else:
--- a/check_sites.py
+++ b/check_sites.py
@@ -21,30 +21,45 @@ from publicsuffix2 import PublicSuffixList
 
 def main():
     args = sys.argv[1:]
-    inputFile = 'first_party_sets.JSON'
-    inputPrefix = ''
-    opts, _ = getopt.getopt(args, "i:", ["data_directory="])
+    input_file = 'first_party_sets.JSON'
+    input_prefix = ''
+    with_diff = False
+    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff="])
     for opt, arg in opts:
         if opt == '-i':
-            inputFile = arg
+            input_file = arg
         if opt == '--data_directory':
-            inputPrefix = arg
+            input_prefix = arg
+        if opt == '--with_diff':
+            with_diff = arg.lower == "true"
 
-    # Open the canonical sites, and load the json
-    with open(inputFile) as f:
+    # Open and load the json of the new list
+    with open(input_file) as f:
         try:
             fps_sites = json.load(f)
         except Exception as inst:
         # If the file cannot be loaded, we will not run any other checks
-            print("There was an error when loading "+ inputFile + 
+            print("There was an error when loading "+ input_file + 
                   "\nerror was: " + inst)
-            exit()      
+            exit()  
+    # Open and load the json of the old list, if there 
+    if with_diff:   
+        with open(os.path.join(input_prefix,'first_party_sets.JSON')) as f:
+            try:
+                prev_sites = json.load(f)
+            except Exception as inst:
+            # If the file cannot be loaded, we will not run any other checks
+                print("There was an error when loading " +
+                    os.path.join(input_prefix,'first_party_sets.JSON') + 
+                    "\nerror was: " + inst)
+                exit()
+     
 
     # Load the etlds from the public suffix list
-    etlds = PublicSuffixList(psl_file = os.path.join(inputPrefix,'effective_tld_names.dat'))
+    etlds = PublicSuffixList(psl_file = os.path.join(input_prefix,'effective_tld_names.dat'))
     # Get all the ICANN domains
     icanns = set()
-    with open(os.path.join(inputPrefix,'ICANN_domains')) as f:
+    with open(os.path.join(input_prefix,'ICANN_domains')) as f:
         for line in f:
             l = line.strip()
             icanns.add(l)
@@ -53,7 +68,7 @@ def main():
     error_texts = []
 
     try:
-        fps_checker.validate_schema(os.path.join(inputPrefix,'SCHEMA.json'))
+        fps_checker.validate_schema(os.path.join(input_prefix,'SCHEMA.json'))
     except Exception as inst:
         # If the schema is invalid, we will not run any other checks
         print(inst)
--- a/check_sites.py
+++ b/check_sites.py
@@ -42,7 +42,7 @@ def main():
         if opt == '--data_directory':
             input_prefix = arg
         if opt == '--with_diff':
-            with_diff = arg.lower == "true"
+            with_diff = arg.lower() == "true"
 
     # Open and load the json of the new list
     with open(input_file) as f:
@@ -89,13 +89,14 @@ def main():
                 exit()
         old_checker = FpsCheck(old_sites, etlds, icanns)
         check_sets, _ = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
-
     else:
         check_sets = fps_checker.load_sets()
     
+    # Run rationale check separately from check_list since it takes no argument
+    fps_checker.has_all_rationales
 
+    # Run rest of checks
     check_list = [
-        fps_checker.has_all_rationales, 
         fps_checker.check_exclusivity,
         fps_checker.find_non_https_urls, 
         fps_checker.find_invalid_eTLD_Plus1,
@@ -110,7 +111,8 @@ def main():
         try:
             check(check_sets)
         except Exception as inst:
-            error_texts.append(inst)
+            error_texts.append("Error while processing" + check + 
+                               "\nError was: " + inst)
     # This message allows us to check the succes of our action
     if fps_checker.error_list or error_texts:
         for checker_error in fps_checker.error_list:
--- a/check_sites.py
+++ b/check_sites.py
@@ -89,8 +89,11 @@ def main():
                 exit()
         old_checker = FpsCheck(old_sites, etlds, icanns)
         check_sets, _ = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
+        # TODO: add variable and check for subtracted_sets in case of user 
+        # removing old set from the list
     else:
         check_sets = fps_checker.load_sets()
+
     
     # Run rationale check separately from check_list since it takes no argument
     fps_checker.has_all_rationales
--- a/check_sites.py
+++ b/check_sites.py
@@ -18,6 +18,17 @@ import sys
 import os
 from publicsuffix2 import PublicSuffixList
 
+def find_diff_sets(old_sets, new_sets):
+    diff_sets = {}
+    subtracted_sets = {}
+    difference = set(old_sets)^set(new_sets)
+    for diff in difference:
+        if diff in new_sets:
+            diff_sets[diff] = new_sets[diff]
+        else:
+            subtracted_sets[diff] = old_sets[diff]
+    return diff_sets, subtracted_sets
+
 
 def main():
     args = sys.argv[1:]
@@ -42,17 +53,6 @@ def main():
             print("There was an error when loading "+ input_file + 
                   "\nerror was: " + inst)
             exit()  
-    # Open and load the json of the old list, if there 
-    if with_diff:   
-        with open(os.path.join(input_prefix,'first_party_sets.JSON')) as f:
-            try:
-                prev_sites = json.load(f)
-            except Exception as inst:
-            # If the file cannot be loaded, we will not run any other checks
-                print("There was an error when loading " +
-                    os.path.join(input_prefix,'first_party_sets.JSON') + 
-                    "\nerror was: " + inst)
-                exit()
      
 
     # Load the etlds from the public suffix list
@@ -75,10 +75,24 @@ def main():
         exit()
 
     check_sets = {}
-    try:
+    # If called with with_diff, we must determine the sets that are different 
+    # to properly construct our check_sets
+    if with_diff:   
+        with open(os.path.join(input_prefix,'first_party_sets.JSON')) as f:
+            try:
+                old_sites = json.load(f)
+            except Exception as inst:
+            # If the file cannot be loaded, we will not run any other checks
+                print("There was an error when loading " +
+                    os.path.join(input_prefix,'first_party_sets.JSON') + 
+                    "\nerror was: " + inst)
+                exit()
+        old_checker = FpsCheck(old_sites, etlds, icanns)
+        check_sets, _ = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
+
+    else:
         check_sets = fps_checker.load_sets()
-    except Exception as inst:
-        error_texts.append(inst)
+    
 
     check_list = [
         fps_checker.has_all_rationales, 
--- a/tests/fps_tests.py
+++ b/tests/fps_tests.py
@@ -1,5 +1,6 @@
 import unittest
 import sys
+import json
 from jsonschema import ValidationError
 from publicsuffix2 import PublicSuffixList
 from unittest import mock
@@ -8,6 +9,7 @@ from requests import structures
 sys.path.append('../first-party-sets')
 from FpsSet import FpsSet
 from FpsCheck import FpsCheck
+from check_sites import find_diff_sets
 
 class TestValidateSchema(unittest.TestCase):
     """A test suite for the validate_schema function of FpsCheck"""
@@ -703,6 +705,40 @@ class TestFindInvalidESLDs(unittest.TestCase):
         }
         self.assertEqual(loaded_sets, expected_sets)
         self.assertEqual(fp.error_list, [])
+
+class TestFindDiff(unittest.TestCase):
+    def test_basic_find_diff(self):
+        old_json_dict = {
+            "sets":
+            [
+                {
+                    "primary": "https://primary.com",
+                    "ccTLDs": {
+                        "https://primary.com": ["https://primary.ca"]
+                    }
+                }
+            ]
+        }
+        new_json_dict = {
+            "sets":
+            [
+                {
+                    "primary": "https://primary2.com",
+                    "ccTLDs": {
+                        "https://primary2.com": ["https://primary2.ca"]
+                    }
+                }
+            ]
+        }
+        old_fp = FpsCheck(fps_sites=old_json_dict,
+                     etlds=None,
+                     icanns=set(["ca"]))
+        new_fp = FpsCheck(fps_sites=new_json_dict,
+                          etlds=None,
+                          icanns=["ca"])
+        diff_sets, _ = find_diff_sets(old_fp.load_sets(), new_fp.load_sets())
+        self.assertEqual(diff_sets, new_fp.load_sets())
+        
 
 # This method will be used in tests below to mock get requests
 def mock_get(*args, **kwargs):
--- a/.github/workflows/fps-submissions-checks.yml
+++ b/.github/workflows/fps-submissions-checks.yml
@@ -32,7 +32,7 @@ jobs:
         run: pip install publicsuffix2
       - name: Content check
         id: check
-        run: python3 main/check_sites.py -i pull-request/first_party_sets.JSON --data_directory main > results.txt
+        run: python3 main/check_sites.py -i pull-request/first_party_sets.JSON --data_directory=main --with_diff=true > results.txt
       - name: Read the result
         id: read_results
         uses: andstor/file-reader-action@v1


check_sites.py
```diff
@@ -18,33 +18,48 @@
...
+
 
 def main():
     args = sys.argv[1:]
-    inputFile = 'first_party_sets.JSON'
-    inputPrefix = ''
-    opts, _ = getopt.getopt(args, "i:", ["data_directory="])
+    input_file = 'first_party_sets.JSON'
+    input_prefix = ''
+    with_diff = False
```
cfredric, line 37: Instead of making the default behavior to validate the whole file, should the default be to validate just the diff, with an opt-out to validate the whole file?

check_sites.py
```diff
@@ -18,33 +18,48 @@
...
+def find_diff_sets(old_sets, new_sets):
+    diff_sets = {}
+    subtracted_sets = {}
+    difference = set(old_sets)^set(new_sets)
+    for diff in difference:
+        if diff in new_sets:
+            diff_sets[diff] = new_sets[diff]
+        else:
+            subtracted_sets[diff] = old_sets[diff]
+    return diff_sets, subtracted_sets
```
cfredric, line 30: It looks like neither callsite (prod, test) use the second return value here. Can we remove it?

check_sites.py
```diff
@@ -18,33 +18,48 @@
 import os
 from publicsuffix2 import PublicSuffixList
 
+def find_diff_sets(old_sets, new_sets):
```
cfredric, line 21: This doesn't find sets that were just modified by the PR, right? It only finds the sets that are completely new or completely removed.

check_sites.py
```diff
@@ -53,20 +68,38 @@ def main():
...
+        # TODO: add variable and check for subtracted_sets in case of user 
+        # removing old set from the list
+    else:
         check_sets = fps_checker.load_sets()
-    except Exception as inst:
-        error_texts.append(inst)
 
+    
+    # Run rationale check separately from check_list since it takes no argument
+    fps_checker.has_all_rationales
```
cfredric, line 99:
```suggestion
    fps_checker.has_all_rationales()
```

cfredric, line 99: Why is this step different? Shouldn't they all either accept the list of sets that they're validating, or all read from the same member variable of sets?

tests/fps_tests.py
```diff
@@ -704,6 +706,40 @@ def test_expected_esld(self):
...
+                    }
+                }
+            ]
+        }
+        old_fp = FpsCheck(fps_sites=old_json_dict,
+                     etlds=None,
+                     icanns=set(["ca"]))
+        new_fp = FpsCheck(fps_sites=new_json_dict,
+                          etlds=None,
+                          icanns=["ca"])
```
cfredric, line 738: I think it would be much clearer if we didn't rely on `FpsCheck` in this unit test, since `find_diff_sets` could easily be tested without specifying all the metadata for FPSes.

I also think it would be clearer if we spelled out the inputs and outputs directly, rather than computing them (via `old_fp.load_sets()` and `new_fp.load_sets()`). See e.g. https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html.

tests/fps_tests.py
```diff
@@ -704,6 +706,40 @@ def test_expected_esld(self):
...
+        diff_sets, _ = find_diff_sets(old_fp.load_sets(), new_fp.load_sets())
+        self.assertEqual(diff_sets, new_fp.load_sets())
```
cfredric, line 740: Can you add more test cases? It would be good to have test cases that show what the expected behavior is for things that are new, modified, or removed.

check_sites.py
```diff
@@ -53,20 +68,38 @@ def main():
...
+    if with_diff:   
+        with open(os.path.join(input_prefix,'first_party_sets.JSON')) as f:
+            try:
+                old_sites = json.load(f)
+            except Exception as inst:
+            # If the file cannot be loaded, we will not run any other checks
+                print("There was an error when loading " +
+                    os.path.join(input_prefix,'first_party_sets.JSON') + 
+                    "\nerror was: " + inst)
+                exit()
```
cfredric, line 89:
```suggestion
                return
```

check_sites.py
```diff
@@ -18,33 +18,48 @@
...
+    with open(input_file) as f:
         try:
             fps_sites = json.load(f)
         except Exception as inst:
         # If the file cannot be loaded, we will not run any other checks
-            print("There was an error when loading "+ inputFile + 
+            print("There was an error when loading "+ input_file + 
                   "\nerror was: " + inst)
-            exit()      
+            exit()
```
cfredric, line 55:
```suggestion
            return  
```

check_sites.py
```diff
@@ -18,33 +18,48 @@
...
 
 def main():
     args = sys.argv[1:]
-    inputFile = 'first_party_sets.JSON'
-    inputPrefix = ''
-    opts, _ = getopt.getopt(args, "i:", ["data_directory="])
+    input_file = 'first_party_sets.JSON'
+    input_prefix = ''
+    with_diff = False
+    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff="])
```
cfredric, line 38:
```suggestion
    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff"])
```

No need to expect an argument for this option, since we can use the option's presence/absence as the bool.

FpsCheck.py
```diff
@@ -103,13 +103,13 @@ def has_all_rationales(self, check_sets):
...
-            check_sets: a dictionary of primary->FpsSet
+            None
         Returns:
             None
         """
         for fpset in self.fps_sites['sets']:
-            sites = check_sets[fpset.get("primary")].associated_sites
-            service_sites = check_sets[fpset.get("primary")].service_sites
+            sites = fpset.get("associatedSites")
+            service_sites = fpset.get("serviceSites")
```
cfredric, line 112: Would this logic (and below) be simpler if we used a default value of `[]` for the associated sites and service sites, if the set doesn't have any? Then you could do:

```python
sites = fpset.get("associatedSites", []) + fpset.get("serviceSites", [])
# ...
if rationales:
  for site in sites:
    # ...
else:
  self.error_list.append(...)
```

FpsCheck.py
```diff
@@ -93,7 +93,7 @@ def load_sets(self):
         self.error_list += load_sets_errors
         return check_sets
 
-    def has_all_rationales(self, check_sets):
+    def has_all_rationales(self):
```
cfredric, line 96: I don't quite follow what this change has to do with the diffs - do we always validate all of the rationales, even on sets that haven't changed?

check_sites.py
sjledoux: You're right. I had meant to use `difference = set(old_sets.items())^set(new_sets.items())`, which should give the correct answer, but is evidently not possible due to FpsSet being unhashable. I'll implement your suggestion below.

check_sites.py
sjledoux: I can remove but I wanted the removed sets variable to be a part of the return statement since we're going to need it anyway when we finalize the policy on removed sets/elements, which I was hoping would happen soon.

cfredric: I'm fine with keeping it as long as we add tests for it :)

check_sites.py
sjledoux: In my opinion, `with_diff` should be `false` by default so that users can easily call `python3 check_sites.py` to run the checks on their local file. Since `input_prefix` is by default set to `''`, setting `with_diff` to `true` by default would result in the checks running on an empty dictionary when someone runs the checks locally.

cfredric: That sounds fine to me :+1:

check_sites.py
sjledoux: Happy to make this change, but is there a specific reason why one would be preferred over the other?

cfredric: I have a slight preference for making the control flow as "local" as possible, just to minimize "spooky action at a distance". But it's not a strong preference here, since this is in the `main` function after all.

check_sites.py
sjledoux: Okay, taken care of now

check_sites.py
sjledoux: Added more tests now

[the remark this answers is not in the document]
sjledoux: Implemented a change that should catch all of the cases in the recent commit

FpsCheck.py
sjledoux: Implemented this in recent commit

check_sites.py
```diff
@@ -18,33 +18,63 @@
...
+        Returns:
+            diff_sets: a dictionary of primary->FpsSet
+            substracted_sets: a dictionary of primary->FpsSet
+    """
+    diff_sets = {}
+    for primary in new_sets:
+        if primary not in old_sets:
+            diff_sets[primary] = new_sets[primary]
+        elif new_sets[primary] != old_sets[primary]:
+            diff_sets[primary] = new_sets[primary]
```
cfredric, line 41: We can simplify the loop and the internal if statement here, with [`dict.items`](https://docs.python.org/3/library/stdtypes.html#dict.items):

```py
diff_sets = {}
for primary, fps in new_sets.items():
    if primary not in old_sets || fps != old_sets[primary]:
        diff_sets[primary] = fps
```

We can then simplify the if statement further, taking advantage of [`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get):

```py
diff_sets = {}
for primary, fps in new_sets.items():
    if fps != old_sets.get(primary):
        diff_sets[primary] = fps
```

I would then probably write this using a [dictionary comprehension](https://www.programiz.com/python-programming/dictionary-comprehension):

```py
diff_sets = {primary: fps
             for primary, fps in new_sets.items()
             if fps != old_sets.get(primary)
            }
```

Declarative descriptions of data are much easier to understand than imperative definitions that build the data over time, IMO.

check_sites.py
```diff
@@ -18,33 +18,63 @@
...
+    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff"])
     for opt, arg in opts:
         if opt == '-i':
-            inputFile = arg
+            input_file = arg
         if opt == '--data_directory':
-            inputPrefix = arg
+            input_prefix = arg
+        if opt == '--with_diff':
+            with_diff = "true"
```
cfredric, line 60:
```suggestion
            with_diff = True
```

FpsCheck.py
sjledoux: This change was made because how has_all_rationales previously worked was causing issues with the change made by implementing the diff changes (it has to do with rationales not being saved as part of the FpsSet object). I reworked the function in the recent commit that fix the way it works with the diff changes, but still take the check_sets argument.

FpsCheck.py
```diff
@@ -103,17 +103,14 @@ def has_all_rationales(self, check_sets):
...
         """
         for fpset in self.fps_sites['sets']:
-            sites = check_sets[fpset.get("primary")].associated_sites
-            service_sites = check_sets[fpset.get("primary")].service_sites
-            if sites:
-                sites = sites + service_sites if service_sites else sites
-            else:
-                sites = service_sites
+            if fpset.get("primary") not in  check_sets:
+                continue
```
cfredric, line 112: Should we just loop over `check_sets` instead? It's not clear to me how/if the data in `self.fps_sites['sets']` is different from `check_sets`, other than being a superset of that dict.

tests/fps_tests.py
```diff
@@ -704,6 +706,184 @@ def test_expected_esld(self):
...
         self.assertEqual(fp.error_list, [])
 
+class TestFindDiff(unittest.TestCase):
+    def test_unchanged_sets(self):
+        old_sets = {
+            'https://primary.com': 
+            FpsSet(
+                    primary="https://primary.com", 
+                    associated_sites=None,
+                    service_sites=None,
```
cfredric, line 716:
```suggestion
```

Same comment elsewhere.

tests/fps_tests.py
```diff
@@ -704,6 +706,184 @@ def test_expected_esld(self):
...
+                    )
+        }
+        diff_sets, subtracted_sets = find_diff_sets(old_sets, new_sets)
+        self.assertEqual(diff_sets, {})
+        self.assertEqual(subtracted_sets, {})
+
+    def test_added_set(self):
+        old_sets = {
+            'https://primary.com': 
+            FpsSet(
```
cfredric, line 740: Since we're relying on `FpsSet.__eq__`, we probably also want to add test cases for that, separately, to ensure that it checks all of the fields, handles `FpsSet` instances that are equivalent but [not the same instance](https://docs.python.org/3/reference/expressions.html#is) correctly (i.e. `fps1 is fps2` is false but `fps1 == fps2` is true), etc.

tests/fps_tests.py
```diff
@@ -704,6 +706,184 @@ def test_expected_esld(self):
...
+                    primary="https://primary2.com", 
+                    associated_sites=None,
+                    service_sites=None,
+                    ccTLDs={
+                        "https://primary2.com": ["https://primary2.ca"]
+                    }
+                    )
+        }
+        expected_sets = {
```
cfredric, line 769:
```suggestion
        expected_diff_sets = {
```

Nit: just because there are a few different collections of sets here, it's probably best if we're specific?

tests/fps_tests.py
```diff
@@ -704,6 +706,184 @@ def test_expected_esld(self):
...
+                    service_sites=None,
+                    ccTLDs={
+                        "https://primary2.com": ["https://primary2.ca"]
+                    }
+                    )
+        }
+        diff_sets, subtracted_sets = find_diff_sets(old_sets, new_sets)
+        self.assertEqual(diff_sets, expected_sets)
+        self.assertEqual(subtracted_sets, {})
+    def test_removed_set(self):
```
cfredric, line 783:
```suggestion
        self.assertEqual(subtracted_sets, {})

    def test_removed_set(self):
```

Please put a blank line between different methods of a class. Same comment elsewhere. https://peps.python.org/pep-0008/#blank-lines

check_sites.py
```diff
@@ -18,33 +18,63 @@
...
+        to create new_sets and returns them as the dictionary diff_sets. 
+        Additionally, finds First-Party Sets that have been removed from
+        old_sets to create new_sets and returns them as subtracted_sets.
+
+        Args:
+            old_sets: a dictionary of primary->FpsSet
+            new_sets: a dictionary of primary->FpsSet
+        Returns:
+            diff_sets: a dictionary of primary->FpsSet
+            substracted_sets: a dictionary of primary->FpsSet
```
cfredric, line 34:
```suggestion
            old_sets: Dict[string, FpsSet]
            new_sets: Dict[string, FpsSet]
        Returns:
            Tuple[Dict[string, FpsSet], Dict[string, FpsSet]]
```

It's probably best if we use python's standard [type hint](https://docs.python.org/3/library/typing.html) syntax.

FpsCheck.py
sjledoux: Since the rationales aren't stored in FpsSet objects, we need to get the rationales from the JSON document stored at `self.fps_sites` when we run `has_all_rationales`. I thought it would be better to iterate through the list of sets stored at `self.fps_sites['sets']` and check for their presence in `check_sets` rather than vice versa because `check_sets` is a dictionary while `self.fps_sites['sets']` is a list. So the current implementation should have time complexity O(N) while doing the alternative would be O(N<sup>2</sup>).

Best practice is probably to re-write the FpsSet class to incorporate the rationales in a variable, but I would like to do that in a future PR when the time comes.

FpsCheck.py
```diff
@@ -103,17 +103,14 @@ def has_all_rationales(self, check_sets):
...
             None
         """
         for fpset in self.fps_sites['sets']:
-            sites = check_sets[fpset.get("primary")].associated_sites
-            service_sites = check_sets[fpset.get("primary")].service_sites
-            if sites:
-                sites = sites + service_sites if service_sites else sites
-            else:
-                sites = service_sites
+            if fpset.get("primary") not in  check_sets:
```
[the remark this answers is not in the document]
Pramualsup22: .github/workflows/fps-submissions-checks.yml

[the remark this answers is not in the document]
Pramualsup22: Ok

check_sites.py
sjledoux: Done

tests/fps_tests.py
sjledoux: Made this change, but this required a change to the FpsSet class in order to work. I will follow up on this in a future PR to make a similar change to ccTLDs, as that might be a little more complicated.

tests/fps_tests.py
sjledoux: Done

tests/fps_tests.py
sjledoux: Done

check_sites.py
sjledoux: Implemented here, but will need to change this in other docstrings in future PR

FpsSet.py
```diff
@@ -25,7 +25,7 @@ class FpsSet:
     relevant_fields_dict: a dictionary mapping the JSON field equivalents
     of each field to their value within the object. 
   """
-    def __init__(self, ccTLDs, primary, associated_sites, service_sites):
+    def __init__(self, ccTLDs, primary, associated_sites=None, service_sites=None):
```
cfredric, line 28:
```suggestion
    def __init__(self, *, ccTLDs, primary, associated_sites=None, service_sites=None):
```

Optional: You could enforce usage of keyword arguments (rather than positional args) by adding `*, ` here.

tests/fps_tests.py
```diff
@@ -704,6 +740,159 @@ def test_expected_esld(self):
...
+        }
+        diff_sets, subtracted_sets = find_diff_sets(old_sets, new_sets)
+        self.assertEqual(diff_sets, new_sets)
+        self.assertEqual(subtracted_sets, {})
+
+
+
+        
+        
+
```
cfredric, line 895:
```suggestion


```
https://peps.python.org/pep-0008/#blank-lines

FpsCheck.py
cfredric: Agreed, I think we ought to make the script rely entirely on the `FpsSet` instances, rather than a mixture of the `FpsSet`s and the raw dictionaries. That can be done in a followup, though.

tests/fps_tests.py
sjledoux: Done

tests/fps_tests.py
```diff
@@ -97,6 +98,47 @@ def test_no_contact(self):
...
+                    primary="https://primary.com")
+        fps_2 = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
+                    }, 
+                    primary="https://primary.com")
+        self.assertFalse(fps_1 is fps_2)
+        self.assertTrue(fps_1 == fps_2)
+        fps_2 = fps_1
+        self.assertTrue(fps_1 is fps_2)
+        self.assertTrue(fps_1 == fps_2)
```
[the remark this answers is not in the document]
sjledoux: Done

tests/fps_tests.py
```diff
@@ -97,6 +98,47 @@ def test_no_contact(self):
...
+        fps_1 = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
+                    }, 
+                    primary="https://primary.com")
+        fps_2 = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
+                    }, 
+                    primary="https://primary.com")
+        self.assertFalse(fps_1 is fps_2)
+        self.assertTrue(fps_1 == fps_2)
```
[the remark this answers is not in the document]
sjledoux: Done

tests/fps_tests.py
```diff
@@ -97,6 +98,47 @@ def test_no_contact(self):
...
+    def test_equal_case(self):
+        fps_1 = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
+                    }, 
+                    primary="https://primary.com")
+        fps_2 = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
+                    }, 
+                    primary="https://primary.com")
+        self.assertFalse(fps_1 is fps_2)
```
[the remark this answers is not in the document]
sjledoux: Done

FpsSet.py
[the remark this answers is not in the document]
sjledoux: Will chew on this and possibly add to future PR.

tests/fps_tests.py
```diff
@@ -1,5 +1,6 @@
 import unittest
 import sys
+import json
```
cfredric, line 3:
```suggestion
```

tests/fps_tests.py
```diff
@@ -704,6 +738,155 @@ def test_expected_esld(self):
...
+            FpsSet(
+                    primary="https://primary2.com",
+                    ccTLDs={
+                        "https://primary2.com": ["https://primary2.ca"]
+                    }
+                    )
+        }
+        diff_sets, subtracted_sets = find_diff_sets(old_sets, new_sets)
+        self.assertEqual(diff_sets, expected_diff_sets)
+        self.assertEqual(subtracted_sets, {})
```
cfredric, line 802:
```suggestion
        self.assertEqual(find_diff_sets(old_sets, new_sets), (expected_diff_sets, {}))
```

Optional: you could test both return values at once, if you wanted to make this a bit more concise.

cfredric (commented):
LGTM % unused import nit

tests/fps_tests.py
```diff
@@ -704,6 +737,153 @@ def test_expected_esld(self):
...
+            FpsSet(
+                    primary="https://primary.com",
+                    ccTLDs={
+                        "https://primary.com": ["https://primary.ca"]
+                    }
+                    )
+        }
+        diff_sets, subtracted_sets = find_diff_sets(old_sets, new_sets)
+        self.assertEqual(diff_sets, {})
+        self.assertEqual(subtracted_sets, {})
```
cfredric, line 762: Same comment here (and in the other tests) about checking both return values at once; I think we should be consistent and do the assertions the same way in each test, unless there's a specific reason we can't (e.g., if we need to assert that `diff_sets` is equal to something, but that `subtracted_sets` is *not* equal to something).

cfredric (commented):
One final test nit.

cfredric (approved):

may251989 (commented):
May251989

may251989:
May251989[](url)

PR #33, opened 2023-06-01
change com exception in find_invalid_alias_eSLDs to look at TLD only

sjledoux:
The `com` exception added to `find_invalid_alias_eSLDs` mistakenly looked at the entire eTLD of the aliased domain, rather than the just the TLD, and checked for its presence in the ICANN list of country code top level domains. Sites that had country code TLDs but with eTLDs not on the list were thus inadvertently excluded from this exception. This meant that `example.ca` could have the alias `example.com` listed in its ccTLDs, but `example.co.uk` could not. This should be amended so that the check looks at just the TLDs when making the exceptions, creating the intended behavior in the check.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -431,24 +431,24 @@ class FpsCheck:
                         "primary, associated site, or service site " +
                         "within the firsty pary set for " + primary)
                 # check the validity of the aliases
-                aliased_domain, aliased_etld = (aliased_site.split(".")[0],
+                aliased_eSLD, aliased_tld = (aliased_site.split(".")[0],
                                                 aliased_site.split(".")[-1])
-                if aliased_etld in self.icanns:
+                if aliased_tld in self.icanns:
                     icann_check = self.icanns.union({"com"})
                 else:
                     icann_check = self.icanns
-                plus1s = [(site, site.split(".")[0], site.split(".")[-1])
+                variants = [(site, site.split(".")[0], site.split(".")[-1])
                             for site in curr_set.ccTLDs[aliased_site]]
-                for eSLD in plus1s:
-                    if eSLD[1] != aliased_domain:
+                for site, eSLD, tld in variants:
+                    if eSLD != aliased_eSLD:
                         self.error_list.append(
                             "The following top level domain must match: " 
                             + aliased_site + ", but is instead: " 
-                            + eSLD[0])
-                    if eSLD[2] not in icann_check:
+                            + site)
+                    if tld not in icann_check:
                         self.error_list.append(
-                            "The provided country code: " + eSLD[2] + 
-                            ", in: " + eSLD[0] + 
+                            "The provided country code: " + tld + 
+                            ", in: " + site + 
                             " is not a ICANN registered country code")
 
     def find_robots_txt(self, check_sets):
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -409,34 +409,35 @@ class FpsCheck:
         """Checks that eSLDs match their alias, and that country codes are 
         members of icann
         Reads the ccTLDs and makes sure that they match their equivalent sites,
-        and that they're domains are part of ICANNS list of country codes.
-        If either of these is not the case, appends an error to the error_list
+        and that their eTLDs are part of ICANN's list of country codes.
+        If either of these is not the case, appends an error to the error_list.
+        Note: A site may list a variant with "com" as its eTLD IFF the site 
+        being aliased has an eTLD on ICANN's list of countrycodes. 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
         for primary in check_sets:
             curr_set = check_sets[primary]
             if curr_set.ccTLDs:
+                # This will be more elegant when the default variables are set to []
+                sites = set([primary] + 
+                            (curr_set.associated_sites if curr_set.associated_sites else []) + 
+                            (curr_set.service_sites if curr_set.service_sites else [])
+                            )
                 for aliased_site in curr_set.ccTLDs:
                     # first check if the aliased site is actually anywhere else
                     # in the fps
-                    if aliased_site != primary:
-                        if curr_set.associated_sites:
-                            sites = curr_set.associated_sites 
-                        else:
-                            sites = []
-                        if curr_set.service_sites:
-                            sites += curr_set.service_sites
-                        if aliased_site not in sites:
-                            self.error_list.append(
-                                "The aliased site " + aliased_site + 
-                                " contained within the ccTLDs must be a " +
-                                "primary, associated site, or service site " +
-                                "within the firsty pary set for " + primary)
+                    if aliased_site not in sites:
+                        self.error_list.append(
+                            "The aliased site " + aliased_site + 
+                            " contained within the ccTLDs must be a " +
+                            "primary, associated site, or service site " +
+                            "within the firsty pary set for " + primary)
                     # check the validity of the aliases
-                    aliased_domain, aliased_tld = aliased_site.split(".", 1)
+                    aliased_domain, aliased_tld = (aliased_site.split(".")[0],
+                                                   aliased_site.split(".")[-1])
                     if aliased_tld in self.icanns:
                         icann_check = self.icanns.union({"com"})
                     else:
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -75,7 +75,7 @@ class FpsCheck:
         Args:
             None
         Returns:
-            a dictionary of string->FpsSet
+            Dict[string, FpsSet]
         """
         check_sets = {}
         load_sets_errors = []
@@ -103,7 +103,7 @@ class FpsCheck:
         error_list
 
         Args:
-            None
+            Dict[string, FpsSet]
         Returns:
             None
         """
@@ -131,7 +131,7 @@ class FpsCheck:
         added to the error_list.
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
@@ -194,7 +194,7 @@ class FpsCheck:
         and appends errors to the error list for any that return false
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
@@ -252,7 +252,7 @@ class FpsCheck:
         and appends errors to the error list for any that return false
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
@@ -348,7 +348,7 @@ class FpsCheck:
         format, or its contents do no match what is expected.
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
@@ -424,7 +424,7 @@ class FpsCheck:
             for aliased_site in curr_set.ccTLDs:
                 # first check if the aliased site is actually anywhere else
                 # in the fps
-                if not curr_set.includes(aliased_site):
+                if not curr_set.includes(aliased_site, False):
                     self.error_list.append(
                         "The aliased site " + aliased_site + 
                         " contained within the ccTLDs must be a " +
@@ -463,7 +463,7 @@ class FpsCheck:
         to the error list.
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
@@ -507,7 +507,7 @@ class FpsCheck:
         does not cause a timeout error. 
 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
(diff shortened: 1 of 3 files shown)


FpsCheck.py
```diff
@@ -409,34 +409,35 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
             None
         """
         for primary in check_sets:
             curr_set = check_sets[primary]
             if curr_set.ccTLDs:
+                # This will be more elegant when the default variables are set to []
+                sites = set([primary] + 
+                            (curr_set.associated_sites if curr_set.associated_sites else []) + 
+                            (curr_set.service_sites if curr_set.service_sites else [])
+                            )
```
cfredric, line 428:
```suggestion
                sites = set([primary, 
                            *(curr_set.associated_sites if curr_set.associated_sites else []),
                            *(curr_set.service_sites if curr_set.service_sites else []),
                            ])
```

Optional: personal preference, but I think this is slightly nicer this way.

FpsCheck.py
```diff
@@ -409,34 +409,35 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
-                                "The aliased site " + aliased_site + 
-                                " contained within the ccTLDs must be a " +
-                                "primary, associated site, or service site " +
-                                "within the firsty pary set for " + primary)
+                    if aliased_site not in sites:
+                        self.error_list.append(
+                            "The aliased site " + aliased_site + 
+                            " contained within the ccTLDs must be a " +
+                            "primary, associated site, or service site " +
+                            "within the firsty pary set for " + primary)
```
cfredric, line 437: Should there be a `continue` after this line? I.e. if `aliased_site` isn't in `sites` then there's probably no point in validating the things that it aliases to, right?

FpsCheck.py
```diff
@@ -409,34 +409,35 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
+        Note: A site may list a variant with "com" as its eTLD IFF the site 
+        being aliased has an eTLD on ICANN's list of countrycodes. 
         Args:
-            check_sets: a dictionary of primary->FpsSet
+            check_sets: Dict[string, FpsSet]
         Returns:
             None
         """
         for primary in check_sets:
             curr_set = check_sets[primary]
```
cfredric, line 422:
```suggestion
        for primary, curr_set in check_sets.items():
```

FYI (optional), there's a simpler pattern for these kinds of lines.

FpsCheck.py
cfredric, line 428: Optional: It might be more readable to add a `includes` method to `FpsSet`, maybe with an optional argument that dictates whether the `ccTLDs` dictionary is checked or not.

FpsCheck.py
```diff
@@ -409,34 +409,35 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
             if curr_set.ccTLDs:
```
cfredric, line 423:
```suggestion
            if not curr_set.ccTLDs:
                continue
```

Optional: you could remove one level of indentation by flipping the sense of this if statement.

FpsCheck.py
sjledoux: My thought was that if a contributor was to make a mistake and include an `aliased_site` that they forgot to add to a subset, misspelled the `aliased_site`, or something else to that effect, we could still inform them of any mistakes in their aliases as well so that they can fix them all in one commit.

FpsCheck.py
sjledoux: Done

FpsCheck.py
sjledoux: Done

FpsCheck.py
sjledoux: Done

FpsCheck.py
```diff
@@ -409,51 +409,47 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
-                            self.error_list.append(
-                                "The following top level domain must match: " 
-                                + aliased_site + ", but is instead: " 
-                                + eSLD[0])
-                        if eSLD[2] not in icann_check:
-                            self.error_list.append(
-                                "The provided country code: " + eSLD[2] + 
-                                ", in: " + eSLD[0] + 
-                                " is not a ICANN registered country code")
+        for primary, curr_set in check_sets.items():
```
cfredric, line 421: Food for thought: the `primary` is also accessible via `curr_set.primary`. So, it seems that it would be fine to accept a `List[FpsSet]`, rather than a `Dict[string, FpsSet]`, since we're not relying on the fact that it's a dictionary.

In general, we should code to the interface we need; we shouldn't impose structure that we don't actually care about (if we can avoid it, anyway).

FpsSet.py
```diff
@@ -34,10 +34,21 @@ def __init__(self, ccTLDs, primary, associated_sites=None, service_sites=None):
...
       if isinstance(obj, FpsSet) and self.primary == obj.primary:
         if self.ccTLDs == obj.ccTLDs:
           if self.associated_sites == obj.associated_sites:
             if self.service_sites == obj.service_sites:
               return True
-      return False
\ No newline at end of file
+      return False
+    
+    def includes(self, domain, with_ccTLDs=False):
```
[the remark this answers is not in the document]
sjledoux: Done.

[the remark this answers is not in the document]
sjledoux: Done.

FpsSet.py
```diff
@@ -34,10 +34,21 @@ def __init__(self, ccTLDs, primary, associated_sites=None, service_sites=None):
...
+       sites = set([self.primary, 
+                        *(self.associated_sites if self.associated_sites else []),
+                        *(self.service_sites if self.service_sites else []),
+                        ])
+       if with_ccTLDs:
+          for _, aliased_sites in self.ccTLDs.items():
+             sites = sites.union(aliased_sites)
+       return domain in sites
```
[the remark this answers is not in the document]
sjledoux: Implemented this suggestion

tests/fps_tests.py
```diff
@@ -703,6 +709,76 @@ def test_valid_com_in_alias(self):
...
+            'https://primary.com': 
+            FpsSet(
+                    primary="https://primary.com",
+                    associated_sites=["https://associated.com"],
+                    ccTLDs={
+                        "https://associated.com": ["https://associated.gov"]
+                    }
+                    )
+        }
+        self.assertEqual(loaded_sets, expected_sets)
```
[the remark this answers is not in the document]
sjledoux: Removed here and elsewhere where it is unnecessary

FpsCheck.py
cfredric: SGTM, I didn't mean to imply it should be done in this PR!

FpsSet.py
```diff
@@ -34,10 +34,20 @@ def __init__(self, ccTLDs, primary, associated_sites=None, service_sites=None):
...
\ No newline at end of file
+      return False
+    
+    def includes(self, domain, with_ccTLDs=True):
+       if (self.primary == domain or
+                   domain in (self.associated_sites or []) or
+                   domain in (self.service_sites or [])):
+           return True
+       if with_ccTLDs:
+           return domain in (variant for variant in self.ccTLDs.values())
```
cfredric, line 52: I don't think this is correct. This generator gives a bunch of lists of domains, but `domain` will never be one of those lists (though it might be *in* one of them). So we need to flatten the lists too.
```suggestion
           return domain in (variant for variant_list in self.ccTLDs.values() for variant in variant_list)
```

Can you add a test case that would have caught this?

tests/fps_tests.py
```diff
@@ -136,6 +136,34 @@ def test_inequal_cases(self):
         fps_2.service_sites = ["https://service2.com"]
         self.assertNotEqual(fps_1, fps_2)
 
+class TestFpsIncludes(unittest.TestCase):
+    def test_primary_case(self):
+        fps = FpsSet(ccTLDs={
+                        "https://primary.com": "https://primary.ca"
```
cfredric, line 142:
```suggestion
                        "https://primary.com": ["https://primary.ca"]
```

The ccTLD dict entries should be lists, shouldn't they? That way there could be multiple different aliases for the same "canonical" domain. Same comment elsewhere.

(Side note - it might be a good idea to add some static type checking in this script, via mypy or something else. Separate PR, of course.)

tests/fps_tests.py
```diff
@@ -296,7 +324,6 @@ def test_servicesets_overlap(self):
                     service_sites=["https://service1.com"]),
```
cfredric, line 325: Looks like `expected_sets` is no longer used here, and probably in other test cases as well.

sjledoux: It's now been removed wherever extraneous. Thank you.

FpsSet.py
sjledoux: You're correct, I was misunderstanding the generators. Now fixed, and should be properly tested.

FpsCheck.py
```diff
@@ -409,51 +409,47 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
+                # in the fps
+                if not curr_set.includes(aliased_site, False):
+                    self.error_list.append(
+                        "The aliased site " + aliased_site + 
+                        " contained within the ccTLDs must be a " +
+                        "primary, associated site, or service site " +
+                        "within the firsty pary set for " + primary)
+                # check the validity of the aliases
+                aliased_domain, aliased_etld = (aliased_site.split(".")[0],
+                                                aliased_site.split(".")[-1])
```
cfredric, line 435: Taking another look at this for the eTLD/TLD change - I'm not sure that this is totally correct. Take for example, the eTLD+1 `cfredric.github.io` (note that `github.io` is on the PSL, so this is really an eTLD+1). For that site, `aliased_site` would be `"https://cfredric.github.io"`, so `aliased_domain` and `aliased_etld` work out to be `"cfredric"` and `"io"`, respectively. But the eTLD should be `"github.io"`, not just `"io"`.

We could adjust the indexing done here in order to get this case right, but it would be thrown off again if I specified the "site" as `"https://www.cfredric.github.io"`. (I'm not sure if we have a check for this kind of input; `FpsCheck.is_eTLD_Plus1` would not catch this.)

I think we need to use the `publicsuffix2` library to really be sure we're using the correct eTLD (up to changes between PSL snapshots, anyway).

sjledoux: I'm not sure I'm fully understanding the issue. This check is meant to ensure that the TLD for a ccTLD variant is actually a country-code, so in the case of `https://cfredric.github.io` all we care is that `io` is a country code recognized by ICANN and therefore `https://cfredric.github.io` is considered a valid variant for another site such as `https://cfredric.com`. As far as I'm aware, this is intended behavior. Should it not be? 

However, I do agree that the `www.` subdomain will throw off how we check the eSLDs, which is something we can try to figure out.

FpsCheck.py
```diff
@@ -409,51 +409,47 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
+                # first check if the aliased site is actually anywhere else
+                # in the fps
+                if not curr_set.includes(aliased_site, False):
+                    self.error_list.append(
+                        "The aliased site " + aliased_site + 
+                        " contained within the ccTLDs must be a " +
+                        "primary, associated site, or service site " +
+                        "within the firsty pary set for " + primary)
+                # check the validity of the aliases
+                aliased_domain, aliased_etld = (aliased_site.split(".")[0],
```
cfredric, line 434:
```suggestion
                aliased_domain, aliased_tld = (aliased_site.split(".")[0],
```

Nit: as discussed offline, this PR doesn't use an eTLD here, it uses the TLD.

cfredric, line 434: I wonder if there's a better word than `domain` we could use here, instead of `aliased_domain`? Strictly speaking, the whole string is the domain, including the TLD.

This code is really just trying to remove the eTLD from the domain. We called the remaining segment the "eSLD" in the submission guidelines, but I don't think that term is correct - the SLD of `example.co.uk` is `co.uk`, not `example` or `example.co`. Maybe the best choice is the most straightforward one: `aliased_site_without_etld` or similar?

I also think we should use the `publicsuffix2` library in order to identify the eTLD, instead of assuming that it's all the domain segments except the first one. But that can be done in a followup.

FpsCheck.py
```diff
@@ -409,51 +409,47 @@ def find_invalid_alias_eSLDs(self, check_sets):
...
+                                                aliased_site.split(".")[-1])
+                if aliased_etld in self.icanns:
+                    icann_check = self.icanns.union({"com"})
+                else:
+                    icann_check = self.icanns
+                plus1s = [(site, site.split(".")[0], site.split(".")[-1])
+                            for site in curr_set.ccTLDs[aliased_site]]
+                for eSLD in plus1s:
```
cfredric, line 442: Optional, FYI: you can destructure loop arguments to give things better names, e.g.:

```suggestion
                for site, site_without_etld, tld in plus1s:
```

FpsCheck.py
cfredric: Resolved offline, the description/title of the PR got the intent backwards, so I misunderstood.

FpsCheck.py
sjledoux: Discussed this offline to resolve. Name changed, but it was decided `eSLD` was an appropriate term in general.

FpsCheck.py
sjledoux: Done

FpsCheck.py
sjledoux: Done

cfredric (approved):

PR #53, opened 2023-07-25
Account for multi-level domains in etld plus 1 check

sjledoux:
Currently, `is_eTLD_Plus1` does not run as intended - it merely confirms that domain names it is passed contain a TLD in the PSL. The function should be changed to properly confirm that the domain does contain a TLD in the PSL, and that it is ONLY that domain plus one more level. For example, `primary.com` and `primary.co.uk` should both be accepted as "com" and "co.uk" are both listed in the PSL. However, `primary.example.com` should not be accepted as "example.com" is not listed in the PSL, therefore `primary.example.com` is two levels higher than a PSL listed domain.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -243,7 +243,9 @@ class FpsCheck:
         Returns:
             boolean with truth value dependent on value of get_public_suffix
         """
-        return self.etlds.get_public_suffix(site, strict=True) is not None
+        ps = self.etlds.get_public_suffix(site, strict=True)
+        return (ps is not None and ps == site)
+    
 
     def find_invalid_eTLD_Plus1(self, check_sets):
         """Checks if all domains are etld+1 compliant
@@ -260,36 +262,34 @@ class FpsCheck:
             # Apply to the primary
             if not self.is_eTLD_Plus1(primary):
                 self.error_list.append(
-                    "The provided primary site does not have an eTLD in the" +
-                    " Public suffix list: " + primary)
+                    "The provided primary site is not an eTLD+1: " +
+                    primary)
             # Apply to the country codes
             if check_sets[primary].ccTLDs:
                 for alias in check_sets[primary].ccTLDs:
                     if not self.is_eTLD_Plus1(alias):
                         self.error_list.append(
-                            "The provided alias does not have an eTLD in the "
-                            + "Public suffix list: " + alias)
+                            "The provided alias is not an eTLD+1: " +
+                            alias)
                     for aliased_site in check_sets[primary].ccTLDs[alias]:
                         if not self.is_eTLD_Plus1(aliased_site):
                             self.error_list.append(
-                                "The provided aliased site does not have an "
-                                + "eTLD in the Public suffix list: " + 
-                                aliased_site)
+                                "The provided aliased site is not an eTLD+1: " 
+                                + aliased_site)
             # Apply to associated sites
             if check_sets[primary].associated_sites:
                 for associated_site in check_sets[primary].associated_sites:
                     if not self.is_eTLD_Plus1(associated_site):
                         self.error_list.append(
-                            "The provided associated site does not have an " +
-                            "eTLD in the Public suffix list: " + 
+                            "The provided associated site is not an eTLD+1: " +
                             associated_site)
             # Apply to service sites
             if check_sets[primary].service_sites:
                 for service_site in check_sets[primary].service_sites:
                     if not self.is_eTLD_Plus1(service_site):
                         self.error_list.append(
-                            "The provided service site does not have an eTLD " 
-                            + "in the Public suffix list: " + service_site)
+                            "The provided service site is not an eTLD+1: " + 
+                            service_site)
 
     def open_and_load_json(self, url):
         """Calls urlopen and returns json from a site
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -244,10 +244,10 @@ class FpsCheck:
             boolean with truth value dependent on value of get_public_suffix
         """
         assert site is not None
-        site = site.lstrip("https://")
-        is_etld_plus1 = self.etlds.get_public_suffix(site, strict=True) == site
-        is_not_etld = self.etlds.get_tld(site, strict=True) != site
-        return is_etld_plus1 and is_not_etld
+        site = site.replace("https://", "", 1)
+        has_etld = self.etlds.get_sld(site, strict=True) == site
+        is_more_than_etld = self.etlds.get_tld(site, strict=True) != site
+        return has_etld and is_more_than_etld
     
 
     def find_invalid_eTLD_Plus1(self, check_sets):
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -243,8 +243,11 @@ class FpsCheck:
         Returns:
             boolean with truth value dependent on value of get_public_suffix
         """
-        ps = self.etlds.get_public_suffix(site, strict=True)
-        return (ps is not None and ps == site)
+        assert site is not None
+        site = site.replace("https://", "")
+        is_etld_plus1 = self.etlds.get_public_suffix(site, strict=True) == site
+        is_not_etld = self.etlds.get_tld(site, strict=True) != site
+        return is_etld_plus1 and is_not_etld
     
 
     def find_invalid_eTLD_Plus1(self, check_sets):
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -245,9 +245,9 @@ class FpsCheck:
         """
         assert site is not None
         site = site.removeprefix("https://")
-        is_etldp1 = self.etlds.get_sld(site, strict=True) == site
+        is_etldp1_or_etld = self.etlds.get_sld(site, strict=True) == site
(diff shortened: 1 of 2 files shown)


tests/fps_tests.py
```diff
@@ -500,16 +498,50 @@ def test_multi_invalid_etlds(self):
...
+            ]
+        }
+        fp = FpsCheck(fps_sites=json_dict,
+                     etlds=PublicSuffixList(
+                        psl_file = 'effective_tld_names.dat'),
+                     icanns=set())
+        loaded_sets = fp.load_sets()
+        fp.find_invalid_eTLD_Plus1(loaded_sets)
+        self.assertEqual(fp.error_list, 
+                         [])
```
cfredric, line 545: Can we have a test case that checks whether a eTLD is an eTLD+1? E.g. trying to submit a set that includes `https://com` or `https://github.io` or something like that.

FpsCheck.py
```diff
@@ -260,36 +262,34 @@ def find_invalid_eTLD_Plus1(self, check_sets):
             # Apply to the primary
             if not self.is_eTLD_Plus1(primary):
```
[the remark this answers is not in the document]
sjledoux: Good idea. This seems like it might pair well with the change to make FpsCheck a non-immutable object.

tests/fps_tests.py
sjledoux: Added test case "test_valid_tld_plus1" for this

tests/fps_tests.py
```diff
@@ -500,16 +498,50 @@ def test_multi_invalid_etlds(self):
...
-          "The provided associated site does not have an eTLD in the" +
-                    " Public suffix list: https://associated1.c2om",
-          "The provided service site does not have an eTLD in the" +
-                    " Public suffix list: https://service1.c2om"])
+         ["The provided primary site is not an eTLD+1: https://primary.c2om",
+          "The provided alias is not an eTLD+1: https://primary.c2om",
+          "The provided aliased site is not an eTLD+1: https://primary.c2om",
+          "The provided associated site is not an eTLD+1: https://associated1.c2om",
+          "The provided service site is not an eTLD+1: https://service1.c2om"])
+    def test_not_etld_plus1(self):
```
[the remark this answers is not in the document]
sjledoux: Done

FpsCheck.py
```diff
@@ -243,7 +243,9 @@ def is_eTLD_Plus1(self, site):
         Returns:
             boolean with truth value dependent on value of get_public_suffix
         """
-        return self.etlds.get_public_suffix(site, strict=True) is not None
+        ps = self.etlds.get_public_suffix(site, strict=True)
+        return (ps is not None and ps == site)
```
cfredric, line 247:
```suggestion
        assert site is not None
        return self.etlds.get_public_suffix(site, strict=True) == site
```
I think this is equivalent and simpler? We should never be passing in `None` as the site, right?

sjledoux: I was wrongly worrying that if ps was `None` that it would throw an error, forgetting that `None == value` is totally valid Python. I'll add this change.

tests/fps_tests.py
sjledoux: I've changed the test, but I had to change the check as well to meet the condition. As it was, the check would work with the `https://com` case but not the `https://github.io` case because of how `get_public_suffix` works. The current version should catch both cases (but I had to do some string manipulation with the "https://" prefix).

FpsCheck.py
```diff
@@ -243,7 +243,12 @@ def is_eTLD_Plus1(self, site):
         Returns:
             boolean with truth value dependent on value of get_public_suffix
         """
-        return self.etlds.get_public_suffix(site, strict=True) is not None
+        assert site is not None
+        site = site.lstrip("https://")
```
cfredric, line 247:
```suggestion
        site = site.removeprefix("https://")
```
[`lstrip`](https://docs.python.org/3/library/stdtypes.html#str.lstrip) isn't what you want here, since it removes all the leading characters from the string if they are in the argument passed in. So, `"https://shhhhh.com".lstrip("https://")` gives `".com"`, rather than the intended `"ptths.com"`.

FpsCheck.py
```diff
@@ -243,7 +243,12 @@ def is_eTLD_Plus1(self, site):
...
+        is_etld_plus1 = self.etlds.get_public_suffix(site, strict=True) == site
```
cfredric, line 248: It looks like we're supposed to use `get_sld` instead of this function: https://github.com/nexB/python-publicsuffix2/blob/4b2a771d621663dba214a66019a91a9989846a93/src/publicsuffix2/__init__.py#L279

FpsCheck.py
```diff
@@ -243,7 +243,12 @@ def is_eTLD_Plus1(self, site):
...
+        is_not_etld = self.etlds.get_tld(site, strict=True) != site
+        return is_etld_plus1 and is_not_etld
```
cfredric, line 250:
```suggestion
        is_etld_plus1 = self.etlds.get_public_suffix(site, strict=True) == site
        is_etld = self.etlds.get_tld(site, strict=True) == site
        return is_etld_plus1 and not is_etld
```

Optional: booleans are often clearer when written in the positive, so that we don't not never use double negatives. (See [go/tott/530](https://goto.google.com/tott/530), I don't think there's a public link.)

tests/fps_tests.py
```diff
@@ -500,16 +499,90 @@ def test_multi_invalid_etlds(self):
...
+            ]
+        }
+        fp = FpsCheck(fps_sites=json_dict,
+                    etlds=PublicSuffixList(
+                        psl_file = 'effective_tld_names.dat'),
+                    icanns=set())
+        loaded_sets = fp.load_sets()
+        fp.find_invalid_eTLD_Plus1(loaded_sets)
+        self.assertEqual(fp.error_list, 
+        ["The provided primary site is not an eTLD+1: https://7.bg"])
```
cfredric, line 585: Nit: indent this line.

FpsCheck.py
sjledoux: Changed it back to `replace`

tests/fps_tests.py
sjledoux: Done.

FpsCheck.py
cfredric: Can we use `removeprefix` instead of `replace`? It's unlikely to be an issue here, but `replace` would remove substrings in the middle of the string, which is never something we want this to do. So `removeprefix` conveys our intent more clearly.

FpsCheck.py
```diff
@@ -243,7 +243,12 @@ def is_eTLD_Plus1(self, site):
         Returns:
             boolean with truth value dependent on value of get_public_suffix
         """
-        return self.etlds.get_public_suffix(site, strict=True) is not None
+        assert site is not None
+        site = site.replace("https://", "", 1)
+        has_etld = self.etlds.get_sld(site, strict=True) == site
+        is_more_than_etld = self.etlds.get_tld(site, strict=True) != site
+        return has_etld and is_more_than_etld
```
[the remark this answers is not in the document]
sjledoux: My thought was thought calling the first variable `is_etldp1` is also a bit confusing since an entry that is itself an effective top level domain would return true for that. I think it might be confusing to mark that an etld is an etld+1, but I'm not sure if there's a better option than what you're suggesting.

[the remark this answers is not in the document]
cfredric: I'm not entirely clear on the semantics of the publicsuffix2 API, but I was guessing that the use of `strict=True` opted out of that behavior and would make the API return `None` if we gave it an eTLD. Can you test it out to be sure?

Alternatively, I think it'd be fine to do something like:

```
is_etldp1_or_etld = self.etlds.get_sld(site, strict=True) == site
is_etld = self.etlds.get_tld(site, strict=True) == site
return is_etldp1_or_etld and not is_etld
```

Since that makes it clearer what the APIs return, and what exactly we're looking for, IMO.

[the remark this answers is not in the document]
sjledoux: I have tested it out, and it doesn't work as intended with slds without the inclusion of the `get_tld` call, as I was referencing [here](https://github.com/GoogleChrome/first-party-sets/pull/53#discussion_r1274114673). 

> I'm not entirely clear on the semantics of the publicsuffix2 API, but I was guessing that the use of `strict=True` opted out of that behavior and would make the API return `None` if we gave it an eTLD. Can you test it out to be sure?

That's what one would think, but the inclusion of `strict=True` is what allows the function to actually refer to the PSL at all. Without the argument, the methods in publicsuffix2 don't do much of anything - they just extract the sld/tld-like section of a url-like string. Here's the relevant part of the description of `get_sld`:

> The logic does not check by default whether the TLD is in the Trie, so for example, 'www.this.local' will return 'this.local'. If you want to ensure the TLD is in the public suffix list, use strict=True.

It's pretty silly IMO.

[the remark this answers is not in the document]
cfredric: Got it, yeah you did mention that there's some overlap between the APIs. +1 that it's not the easiest to use, but I think we're using it correctly now at least.

cfredric (approved):

PR #59, opened 2023-08-02
Add check for subtracted sets

sjledoux:
Currently, the workflow does not run a check for a case where a pull request removes an entire set from the list. A check should therefore be added ensuring that the well-known page for the primary of a removed set serves an error 404. This will confirm that the set was intended to be removed from the list by the owner of the affected domain.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -425,11 +425,11 @@ class FpsCheck:
                 r = requests.get(url, timeout=10)
                 if r.status_code != 404:
                     self.error_list.append("The set associated with " + primary
-                            + " was removed from the list, but " + primary + WELL_KNOWN + 
+                            + " was removed from the list, but " + url + 
                             " does not return error 404.")
             except Exception as inst:
                 self.error_list.append("Unexpected error when accessing " +
-                                    primary + "; Received error:" + str(inst))
+                                    url + "; Received error:" + str(inst))
 
     def find_invalid_alias_eSLDs(self, check_sets):
         """Checks that eSLDs match their alias, and that country codes are 
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -19,6 +19,7 @@ from urllib.request import urlopen
 from urllib.request import Request
 from publicsuffix2 import PublicSuffixList
 
+WELL_KNOWN = "/.well-known/first-party-set.json"
 
 class FpsCheck:
 
@@ -35,7 +36,7 @@ class FpsCheck:
                 without any given check failing halfway through and not 
                 catching other issues. 
   """
-    WELLKNOWN = "/.well-known/first-party-set.json"
+    
 
     def __init__(self, fps_sites: json, etlds: PublicSuffixList, icanns: set):
         """Stores the input from canonical_sites, effective_tld_names.dat, and 
@@ -324,7 +325,7 @@ class FpsCheck:
             None
         """
         for site in site_list:
-            url = site + self.WELLKNOWN
+            url = site + WELL_KNOWN
             try:
                 json_schema = self.open_and_load_json(url)
                 if 'primary' not in json_schema.keys():
@@ -359,7 +360,7 @@ class FpsCheck:
         # Check the schema to ensure consistency
         for primary in check_sets:
             # First we check the primary sites
-            url = primary + self.WELLKNOWN
+            url = primary + WELL_KNOWN
             # Read the well-known files and check them against the schema we 
             # have stored
             try:
@@ -419,7 +420,7 @@ class FpsCheck:
         Returns:
             None"""
         for primary in subtracted_sets:
-            url = primary + self.WELLKNOWN
+            url = primary + WELL_KNOWN
             try:
                 r = requests.get(url, timeout=10)
                 if r.status_code != 404:
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -35,6 +35,7 @@ class FpsCheck:
                 without any given check failing halfway through and not 
                 catching other issues. 
   """
+    WELLKNOWN = "/.well-known/first-party-set.json"
 
     def __init__(self, fps_sites: json, etlds: PublicSuffixList, icanns: set):
         """Stores the input from canonical_sites, effective_tld_names.dat, and 
@@ -323,7 +324,7 @@ class FpsCheck:
             None
         """
         for site in site_list:
-            url = site + "/.well-known/first-party-set.json"
+            url = site + self.WELLKNOWN
             try:
                 json_schema = self.open_and_load_json(url)
                 if 'primary' not in json_schema.keys():
@@ -358,7 +359,7 @@ class FpsCheck:
         # Check the schema to ensure consistency
         for primary in check_sets:
             # First we check the primary sites
-            url = primary + "/.well-known/first-party-set.json"
+            url = primary + self.WELLKNOWN
             # Read the well-known files and check them against the schema we 
             # have stored
             try:
@@ -418,7 +419,7 @@ class FpsCheck:
         Returns:
             None"""
         for primary in subtracted_sets:
-            url = primary + "/.well-known/first-party-set.json"
+            url = primary + self.WELLKNOWN
             try:
                 r = requests.get(url, timeout=10)
                 if r.status_code != 404:
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -425,7 +425,7 @@ class FpsCheck:
                 r = requests.get(url, timeout=10)
                 if r.status_code != 404:
                     self.error_list.append("The set associated with " + primary
-                            + " was removed from the list, but " + primary + 
+                            + " was removed from the list, but " + primary + WELL_KNOWN + 
                             " does not return error 404.")
             except Exception as inst:
                 self.error_list.append("Unexpected error when accessing " +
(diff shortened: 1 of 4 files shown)


check_sites.py
```diff
@@ -100,16 +111,19 @@ def main():
...
-        check_sets, _ = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
+        check_sets, subtracted_sets = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
         # TODO: add variable and check for subtracted_sets in case of user 
         # removing old set from the list
     else:
         check_sets = fps_checker.load_sets()
 
+    # Run check on subtracted sets
+    if subtracted_sets:
+        fps_checker.find_invalid_removal(subtracted_sets)
```
cfredric, line 122:
```suggestion
    fps_checker.find_invalid_removal(subtracted_sets)
```

The check itself should be a no-op if the dict arg is empty, so we don't have to skip it in that case. It makes things clearer if the caller code can be straight-line, and rely on the checks themselves to do as much or little work as they need to.

check_sites.py
```diff
@@ -85,8 +88,16 @@ def main():
         # If the schema is invalid, we will not run any other checks
         print(inst)
         return
+    
+    # Check for exclusivity among all sets in the updated version
+    try:
+        fps_checker.check_exclusivity(fps_checker.load_sets())
+    except Exception as inst:
+            error_texts.append(inst)
```
cfredric, line 96: Why are we moving this check from its original location in the list of checks?

check_sites.py
```diff
@@ -100,16 +111,19 @@ def main():
                     "\nerror was: " + inst)
                 return
         old_checker = FpsCheck(old_sites, etlds, icanns)
-        check_sets, _ = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
+        check_sets, subtracted_sets = find_diff_sets(old_checker.load_sets(), fps_checker.load_sets())
         # TODO: add variable and check for subtracted_sets in case of user 
         # removing old set from the list
     else:
         check_sets = fps_checker.load_sets()
```
cfredric, line 118: We also need to assign `subtracted_sets` here, right? This is the case when we're checking the entire list's invariants (not just the diff), which means we ought to consider what was removed too.

check_sites.py
```diff
@@ -36,9 +36,12 @@ def find_diff_sets(old_sets, new_sets):
...
              if fps != old_sets.get(primary)
             }
-
     subtracted_sets = {primary: old_sets[primary] for primary in set(old_sets) - set(new_sets)}
-
+    # Make sure that subtracted sets aren't just sets with changed primaries
+    for _, fps in diff_sets.items():
+        for primary in subtracted_sets:
+            if fps.includes(primary):
+                subtracted_sets -= primary
```
cfredric, line 44: Can you write a test that would fail if we didn't have this code?

cfredric, line 44: I think a simpler way of handling this scenario would be to include it in the initial definition of `subtracted_sets` like this:

```suggestion
    subtracted_sets = {
        primary: old_sets[primary]
        for primary in set(old_sets) - set(new_sets)
        if primary not in new_sets
    }
```

cfredric: Oh, I guess we'd have to make the guard call `fps.includes`:

```suggestion
    subtracted_sets = {
        primary: old_sets[primary]
        for primary in set(old_sets) - set(new_sets)
        if not any(fps.includes(primary) for fps in new_sets)
    }
```

check_sites.py
sjledoux: I don't think so? In this case, there is no sense of an old list so there is nothing to check against to determine if an element was removed

FpsCheck.py
```diff
@@ -407,6 +407,27 @@ def find_invalid_well_known(self, check_sets):
                 for aliased_site in check_sets[primary].ccTLDs:
                     ccTLD_sites += check_sets[primary].ccTLDs[aliased_site]
                     self.check_list_sites(primary, ccTLD_sites)
+        
+    def find_invalid_removal(self, subtracted_sets):
+        """Checks that any sets being removed were properly removed by owner
+        
+        Checks that the /.well-known page for the primary of any FPS removed
```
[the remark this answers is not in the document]
sjledoux: That isn't the policy that we landed on. We thought that just the primary serving the 404 should be enough to determine that set removal was intended by the owner. Is there reason to believe that it wouldn't be?

check_sites.py
sjledoux: As I was editing this I realized that keeping it within the `with_diff` portion meant that the check was only looking at the exclusivity within the new sets portion, essentially defeating the purpose of the check.

cfredric: Ah, great point!

cfredric: Can you write a test case that exercises this? I.e. one that ought to fail the checks, but would pass if you hadn't moved this one.

FpsCheck.py
```diff
@@ -35,6 +35,7 @@ class FpsCheck:
                 without any given check failing halfway through and not 
                 catching other issues. 
   """
+    WELLKNOWN = "/.well-known/first-party-set.json"
```
cfredric, line 38: If nothing outside of this file needs to use this string (e.g. no tests -- do they also repeat the well-known path?), I would make it a file-local constant, instead of a class-scoped constant. That reduces the visibility of it, which enforces more encapsulation.

cfredric, line 38:
```suggestion
    WELL_KNOWN = "/.well-known/first-party-set.json"
```
Nit: Python typically names things in CONSTANT_CASE using underscores to separate words. See https://peps.python.org/pep-0008/#constants

check_sites.py
sjledoux: In case it isn't clear, [this test](https://github.com/GoogleChrome/first-party-sets/pull/59/files#diff-c40416e23022656073da5094bfddfab89ac7e41d556203fd1968b7135308f365R872) that I added after your comment is meant to address the case.

sjledoux: Done

check_sites.py
sjledoux: I suppose I can do that, but since this is a change in `check_sites.py` I think I would have to create a new test file that executes the script with different input files (or spoofed versions of input files)? That might make this PR a bit longer. I understand wanting to have the change explicitly tested, and I did test it locally with made up files, just not sure if there's a better way to create an automated test for this type of change?

cfredric: Does `check_sets.py` have to be the thing that knows how to find/parse the input files *and* knows which checks to run? I think we could rearrange things so that check_sets.py is responsible for reading/parsing the input file(s) and CLI options, but then hands the parsed input off to something else to actually do the validation and create the appropriate errors. (And that way we could have unit tests for all of the validation steps, and the tests wouldn't have to worry about input files - they'd just supply the desired hardcoded inputs to the thing that does the validation.)

I agree with you though that that seems like a bigger refactoring and probably doesn't belong in this CL. (Strictly speaking, each CL/PR should make only one change, so fixing the disjointness check belongs in a separate PR. But it's already written, and small, so I won't make you move it now.)

check_sites.py
```diff
@@ -36,9 +36,11 @@ def find_diff_sets(old_sets, new_sets):
              for primary, fps in new_sets.items()
              if fps != old_sets.get(primary)
             }
-
-    subtracted_sets = {primary: old_sets[primary] for primary in set(old_sets) - set(new_sets)}
-
+    subtracted_sets = {
+        primary: old_sets[primary]
+        for primary in set(old_sets) - set(new_sets)
+        if not any(fps.includes(primary) for _, fps in new_sets.items())
```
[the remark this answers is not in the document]
sjledoux: Done.

cfredric (commented):
LGTM % one nit

FpsCheck.py
```diff
@@ -407,6 +409,27 @@ def find_invalid_well_known(self, check_sets):
...
+            subtracted_sets: Dict[string, FpsSet]
+        Returns:
+            None"""
+        for primary in subtracted_sets:
+            url = primary + WELL_KNOWN
+            try:
+                r = requests.get(url, timeout=10)
+                if r.status_code != 404:
+                    self.error_list.append("The set associated with " + primary
+                            + " was removed from the list, but " + primary + WELL_KNOWN +
```
cfredric, line 428:
```suggestion
                            + " was removed from the list, but " + url + 
```

On second look, this way would be cleaner.

tests/fps_tests.py
```diff
@@ -1124,7 +1154,40 @@ def test_404_redirect(self, mock_get):
...
+                    ccTLDs={}
+                    )
+        }
+        fp = FpsCheck(fps_sites={},
+                     etlds=None,
+                     icanns=set())
+        fp.find_invalid_removal(subtracted_sets)
+        self.assertEqual(fp.error_list, ["The set associated with " +
+                "https://primary1.com was removed from the list, but " +
+                "https://primary1.com does not return error 404."])
```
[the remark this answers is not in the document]
sjledoux: Oops. That's what I get for using the "commit suggestion" button without testing the change.

first_party_sets.JSON
cfredric, line 1: Please revert this - we should not change `first_party_sets.JSON` as part of this PR. (IIUC this change came from another commit that was already merged on main?)

cfredric (approved):
LGTM, thanks!

PR #69, opened 2023-09-01
Fix well-known check to match fields of FPS entry with JSON file

sjledoux:
Currently, if the well-known page of an FPS entry's primary contains a field that is not in the FPS entry itself, a NoneType Error will occur during the well-known check, and this will surface to the uploader when they attempt to make their PR, such as here as can be seen [here](https://github.com/GoogleChrome/first-party-sets/pull/67#issuecomment-1694200106). This error does not cause any technical issues with the FPS upload process, but it should be changed so that this case is properly caught and communicated to the uploader, allowing them to understand how their PR should be fixed in order to pass the checks and be merged.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -330,7 +330,7 @@ class FpsCheck:
                     self.error_list.append(
                         "The listed associated site site did not have primary"
                         + " as a key in its " + WELL_KNOWN
-                        + "file: " + site)
+                        + " file: " + site)
                 elif json_schema['primary'] != primary:
                     self.error_list.append("The listed associated site "
                     + "did not have " + primary + " listed as its primary: " 
@@ -341,11 +341,24 @@ class FpsCheck:
                     + "error was: " + str(inst))
     
     def check_well_known_list(self, field, list1, list2):
-        """Checks that 2 lists for a given field match each other"""
+        """Checks that 2 lists for a given field match each other
+        
+        Applies a symmetric diff to list1 and list2 and returns an empty list
+        if the 2 fields are symmetric. Otherwise returns a list with a single
+        string to be used as error text, containing the field, list1, list2, 
+        and their symmetric diff.
+
+        Args:
+            field: string
+            list1: list[string]
+            list2: list[string]
+        Returns:
+            list[string]
+        """
         if list1 == list2:
             return []
         diff = sorted(set(list1) ^ set(list2))
-        return ["Experienced a mismatch between the PR submission and the " 
+        return ["Encountered an inequality between the PR submission and the " 
         + WELL_KNOWN + " file:\n\t" + field + " was " + str(list1) + " in the PR, and "
         + str(list2) + " in the well-known.\n\tDiff was: " + str(diff) + "."]
 
@@ -379,37 +392,28 @@ class FpsCheck:
                     json_schema.get('associatedSites'), 
                     json_schema.get('serviceSites'))
                 if well_known_set.primary != curr_fps_set.primary:
-                    field_sym_difference = [well_known_set.primary, curr_fps_set.primary]
-                    self.error_list.append("The following primary was not "
-                    + "present in both the PR and " + WELL_KNOWN + " file: "
-                    + str(sorted(field_sym_difference)))
-                associated_diff = self.check_well_known_list(
+                    self.error_list.append("The " + WELL_KNOWN + " set's " + 
+                    "primary (" + well_known_set.primary + ") did not equal " +
+                    "the PR set's primary (" + curr_fps_set.primary + ")")
+                self.error_list.extend(self.check_well_known_list(
                     "associatedSites",
                     curr_fps_set.associated_sites, 
-                    well_known_set.associated_sites)
-                self.error_list.extend(associated_diff)
-                service_diff = self.check_well_known_list(
+                    well_known_set.associated_sites
+                    )
+                )
+                self.error_list.extend(self.check_well_known_list(
                     "serviceSites",
                     curr_fps_set.service_sites, 
-                    well_known_set.service_sites)
-                self.error_list.extend(service_diff)
-                aliased_site_diff = self.check_well_known_list(
-                    "ccTLD key list", 
-                    curr_fps_set.ccTLDs.keys(), 
-                    well_known_set.ccTLDs.keys())
-                self.error_list.extend(aliased_site_diff)
+                    well_known_set.service_sites
+                    )
+                )
                 for aliased_site in curr_fps_set.ccTLDs | well_known_set.ccTLDs:
-                    list1 = []
-                    if aliased_site in curr_fps_set.ccTLDs:
-                        list1 = curr_fps_set.ccTLDs[aliased_site]
-                    list2 = []
-                    if aliased_site in well_known_set.ccTLDs:
-                        list2 = well_known_set.ccTLDs[aliased_site]
-                    alias_diff = self.check_well_known_list(
+                    self.error_list.extend(self.check_well_known_list(
                         aliased_site + " alias list",
-                        list1,
-                        list2)
-                    self.error_list.extend(alias_diff)
+                        curr_fps_set.ccTLDs.get(aliased_site, []),
+                        well_known_set.ccTLDs.get(aliased_site, [])
+                        )
+                    )
             except Exception as inst:
                 self.error_list.append(
                     "Experienced an error when trying to access " + url + 
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -329,7 +329,7 @@ class FpsCheck:
                 if 'primary' not in json_schema.keys():
                     self.error_list.append(
                         "The listed associated site site did not have primary"
-                        + " as a key in its .well-known/first-party-set.json "
+                        + " as a key in its " + WELL_KNOWN
                         + "file: " + site)
                 elif json_schema['primary'] != primary:
                     self.error_list.append("The listed associated site "
@@ -339,7 +339,16 @@ class FpsCheck:
                 self.error_list.append(
                     "Experienced an error when trying to access " + url + "; "
                     + "error was: " + str(inst))
-                
+    
+    def check_well_known_list(self, field, list1, list2):
+        """Checks that 2 lists for a given field match each other"""
+        if list1 == list2:
+            return []
+        diff = sorted(set(list1) ^ set(list2))
+        return ["Experienced a mismatch between the PR submission and the " 
+        + WELL_KNOWN + " file:\n\t" + field + " was " + str(list1) + " in the PR, and "
+        + str(list2) + " in the well-known.\n\tDiff was: " + str(diff) + "."]
+
     def find_invalid_well_known(self, check_sets):
         """Checks for and validates well-known pages for FPS sets
 
@@ -372,40 +381,35 @@ class FpsCheck:
                 if well_known_set.primary != curr_fps_set.primary:
                     field_sym_difference = [well_known_set.primary, curr_fps_set.primary]
                     self.error_list.append("The following primary was not "
-                    + "present in both the changelist and "
-                    + ".well-known/first-party-set.json file: "
+                    + "present in both the PR and " + WELL_KNOWN + " file: "
                     + str(sorted(field_sym_difference)))
-                if well_known_set.associated_sites != curr_fps_set.associated_sites:
-                    field_sym_difference = set(
-                       well_known_set.associated_sites) ^ set(curr_fps_set.associated_sites)
-                    self.error_list.append("The following member(s) of " 
-                    + "associatedSites were not present in both the "
-                    + "changelist and .well-known/first-party-set.json "
-                    + "file: " + str(sorted(field_sym_difference)))
-                if well_known_set.service_sites != curr_fps_set.service_sites:
-                    field_sym_difference = set(
-                       well_known_set.service_sites) ^ set(curr_fps_set.service_sites)
-                    self.error_list.append("The following member(s) of " 
-                    + "serviceSites were not present in both the "
-                    + "changelist and .well-known/first-party-set.json "
-                    + "file: " + str(sorted(field_sym_difference)))
-                if well_known_set.ccTLDs != curr_fps_set.ccTLDs:
-                    field_sym_difference = set(
-                        well_known_set.ccTLDS.keys()) ^ set(
-                            curr_fps_set.ccTLDs.keys())
-                    self.error_list.append("The following aliased ccTLD sites"
-                    + "were not present in both the changelist and " 
-                    + ".well-known/first-party-set.json file: "
-                    + str(sorted(field_sym_difference)))
-                    for aliased_site in curr_fps_set.ccTLDs & well_known_set.ccTLDs:
-                        field_sym_difference = set(
-                        well_known_set.ccTLDS[aliased_site]) ^ set(
-                            curr_fps_set.ccTLDs[aliased_site])
-                        self.error_list.append("The following ccTLD aliases"
-                        + "were not present in both the changelist and " 
-                        + ".well-known/first-party-set.json file for site "
-                        + aliased_site + ": " 
-                        + str(sorted(field_sym_difference)))
+                associated_diff = self.check_well_known_list(
+                    "associatedSites",
+                    curr_fps_set.associated_sites, 
+                    well_known_set.associated_sites)
+                self.error_list.extend(associated_diff)
+                service_diff = self.check_well_known_list(
+                    "serviceSites",
+                    curr_fps_set.service_sites, 
+                    well_known_set.service_sites)
+                self.error_list.extend(service_diff)
+                aliased_site_diff = self.check_well_known_list(
+                    "ccTLD key list", 
+                    curr_fps_set.ccTLDs.keys(), 
+                    well_known_set.ccTLDs.keys())
+                self.error_list.extend(aliased_site_diff)
+                for aliased_site in curr_fps_set.ccTLDs | well_known_set.ccTLDs:
+                    list1 = []
+                    if aliased_site in curr_fps_set.ccTLDs:
+                        list1 = curr_fps_set.ccTLDs[aliased_site]
+                    list2 = []
+                    if aliased_site in well_known_set.ccTLDs:
+                        list2 = well_known_set.ccTLDs[aliased_site]
+                    alias_diff = self.check_well_known_list(
+                        aliased_site + " alias list",
+                        list1,
+                        list2)
+                    self.error_list.extend(alias_diff)
             except Exception as inst:
                 self.error_list.append(
                     "Experienced an error when trying to access " + url + 
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -83,9 +83,9 @@ class FpsCheck:
         load_sets_errors = []
         for fpset in self.fps_sites['sets']:
             primary = fpset.get('primary', None)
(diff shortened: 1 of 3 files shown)


FpsSet.py
```diff
@@ -25,7 +25,7 @@ class FpsSet:
     relevant_fields_dict: a dictionary mapping the JSON field equivalents
     of each field to their value within the object. 
   """
-    def __init__(self, ccTLDs, primary, associated_sites=None, service_sites=None):
+    def __init__(self, ccTLDs, primary, associated_sites=[], service_sites=[]):
         self.ccTLDs = ccTLDs
         self.primary = primary
```
cfredric, line 30: Default args only override the values of arguments when no arg was passed in (e.g. `foo()` instead of `foo(x)`). So if we explicitly pass in `None` (or a variable whose value is `None`), we'll still end up storing `None` in the class members.

We also have the same issue with `ccTLDs` as for the other args, so IMO we should handle default values for that too while we're at it.

We could do something like this:

```suggestion
    def __init__(self, ccTLDs, primary, associated_sites=[], service_sites=[]):
        if ccTLDs is None:
            ccTLDs = {}
        if associated_sites is None:
            associated_sites = []
        if service_sites is None:
            service_sites = []
        self.ccTLDs = ccTLDs
        self.primary = primary
```

This way, it's much harder for callers to use an FpsSet instance in a way that's reasonable but would still break.

FpsCheck.py
```diff
@@ -83,9 +83,9 @@ def load_sets(self):
         load_sets_errors = []
         for fpset in self.fps_sites['sets']:
             primary = fpset.get('primary', None)
-            ccTLDs = fpset.get('ccTLDs', None)
-            associated_sites = fpset.get('associatedSites', None)
-            service_sites = fpset.get('serviceSites', None)
+            ccTLDs = fpset.get('ccTLDs', {})
+            associated_sites = fpset.get('associatedSites', [])
+            service_sites = fpset.get('serviceSites', [])
```
cfredric, line 88: IMO the defaults should not be handled here; they should be handled more centrally, in the `FpsSet` constructor.

```suggestion
            ccTLDs = fpset.get('ccTLDs')
            associated_sites = fpset.get('associatedSites')
            service_sites = fpset.get('serviceSites')
```

FpsCheck.py
```diff
@@ -365,9 +365,11 @@ def find_invalid_well_known(self, check_sets):
             # have stored
             try:
                 json_schema = self.open_and_load_json(url)
-                schema_fields = set(self.acceptable_fields) & set(
-                    json_schema.keys())
                 curr_fps_set = check_sets[primary]
+                schema_fields = set({field for field in json_schema.keys()
+                                     if field in self.acceptable_fields})
+                schema_fields.update({field for field in self.acceptable_fields
+                        if len(curr_fps_set.relevant_fields_dict[field]) > 0})
```
cfredric, line 372: Couple of suggestions here:
1. The `{...}` expression is already a [set comprehension](https://python-reference.readthedocs.io/en/latest/docs/comprehensions/set_comprehension.html), so you don't need to wrap it in `set(...)`.
2. Instead of defining `schema_fields` and then immediately mutating it, we should just define it properly the first time, using a more detailed set comprehension.
3. It's idiomatic in Python to check if a list is empty by implicitly converting it to bool, rather than by checking `len(x) > 0`.

```suggestion
                schema_fields = {
                    field
                    for field in self.acceptable_fields
                    if field in json_schema or curr_fps.set.relevant_fields_dict[field]
                }
```

FpsCheck.py
```diff
@@ -376,8 +378,9 @@ def find_invalid_well_known(self, check_sets):
                         else:
                             field_sym_difference = []
                     else:
-                        field_sym_difference = set(json_schema[field]) ^ set(
-                        curr_fps_set.relevant_fields_dict[field])
+                        field_sym_difference = set(json_schema[field] 
+                                if field in json_schema.keys() else []) ^ set(
```
cfredric, line 382: Hm, instead of handling defaults again here, WDYT about converting the JSON object into an `FpsSet` instance? That would discard irrelevant fields for us, and let this benefit from all the default fields we're setting up there. I think that would be less error-prone than manipulating the JSON directly.

FpsCheck.py
```diff
@@ -365,9 +365,11 @@ def find_invalid_well_known(self, check_sets):
             # have stored
             try:
                 json_schema = self.open_and_load_json(url)
-                schema_fields = set(self.acceptable_fields) & set(
-                    json_schema.keys())
                 curr_fps_set = check_sets[primary]
+                schema_fields = set({field for field in json_schema.keys()
+                                     if field in self.acceptable_fields})
```
cfredric, line 370: FYI: since `self.acceptable_fields` is always the same for every FpsCheck instance, and it is never modified, it really should be a file-level constant (i.e. `ACCEPTABLE_FIELDS`), instead of a class member variable.

tests/fps_tests.py
```diff
@@ -1273,6 +1273,69 @@ def test_expected_case(self, mock_open_and_load_json):
         loaded_sets = fp.load_sets()
         fp.find_invalid_well_known(loaded_sets)
         self.assertEqual(fp.error_list, [])
+    @mock.patch('FpsCheck.FpsCheck.open_and_load_json',
```
cfredric, line 1276: Missing blank line between this test and the previous one. Same comment for the other new tests.

FpsCheck.py
cfredric, line 382:
```suggestion
                        field_sym_difference = set(json_schema.get(field, [])) ^ set(
```

tests/fps_tests.py
```diff
@@ -1273,6 +1273,69 @@ def test_expected_case(self, mock_open_and_load_json):
...
+                    }
+                }
+            ]
+        }
+        fp = FpsCheck(fps_sites=json_dict,
+                     etlds=None,
+                     icanns=set())
+        loaded_sets = fp.load_sets()
+        fp.find_invalid_well_known(loaded_sets)
+        self.assertEqual(sorted(fp.error_list), [])
```
cfredric, line 1338: Maybe add a test where the .well-known file has an unrecognized/unvalidated field?

FpsCheck.py
cfredric: Just occurred to me - `field` might be `"ccTLDs"`, and in that case, `[]` isn't a valid value for that field, right? We'd want to use `{}` instead. Similarly, when `field` is `"primary"`, `[]` isn't valid.

But I think in general it might be easier/clearer if we didn't try to be generic over the field, and just check the four fields explicitly. Two of the four fields share the same type, but the 4 fields have 3 different types overall (`str`, `List[str]`, `Dict[str, str]`).

sjledoux: I am open to converting the JSON object into an FPS instance. In terms of your comment about using `{}` vs `[]`, I don't believe that this should matter. Regardless of what the default value is - `[]`, `{}`, `""`, etc. - it's placed inside a set constructor so it will ultimately be equal to an empty set if the field is not present. Primary is already handled differently, but this is because placing a string in a set constructor does a very silly thing.

FpsCheck.py
```diff
@@ -366,33 +364,48 @@ def find_invalid_well_known(self, check_sets):
...
+                    json_schema.get('serviceSites'))
+                if well_known_set.primary != curr_fps_set.primary:
+                    field_sym_difference = [well_known_set.primary, curr_fps_set.primary]
+                    self.error_list.append("The following primary was not "
+                    + "present in both the changelist and "
+                    + ".well-known/first-party-set.json file: "
+                    + str(sorted(field_sym_difference)))
+                if well_known_set.associated_sites != curr_fps_set.associated_sites:
+                    field_sym_difference = set(
+                       well_known_set.associated_sites) ^ set(curr_fps_set.associated_sites)
```
cfredric, line 380: The fact that we're checking equality of the _lists_, but only displaying the symmetric difference between the _sets_ means that comparing `['a', 'b']` and `['b', 'a']` would give an error like `"The following member(s) of associatedSites were not present in both the changelist and .well-known/first-party-set.json file: []"`. That seems like it would be pretty hard to debug.

Ideally, we would print just the necessary and sufficient information needed to fix the issue in the PR (possibly by computing the Levenshtein distance and the corresponding edits between the two lists?). But in the absence of that, I think it would be better to print the two input lists themselves, and also print the difference (either as a symmetric difference, or as both asymmetric differences). Then the error message would contain enough to diagnose what the issue is, without digging into the repo's code - even though it's more verbose. WDYT?

FWIW, I do think it's good to be sensitive to the order that the sites are listed in the subsets, since Chrome has a limit for the number of associated sites it accepts into a set. Enforcing that the .well-known files declare sites in the same order as the GitHub repo does will help ensure consistency of the ordering between the different set declarations.

FpsCheck.py
```diff
@@ -366,33 +364,48 @@ def find_invalid_well_known(self, check_sets):
...
+                    + "associatedSites were not present in both the "
+                    + "changelist and .well-known/first-party-set.json "
+                    + "file: " + str(sorted(field_sym_difference)))
+                if well_known_set.service_sites != curr_fps_set.service_sites:
+                    field_sym_difference = set(
+                       well_known_set.service_sites) ^ set(curr_fps_set.service_sites)
+                    self.error_list.append("The following member(s) of " 
+                    + "serviceSites were not present in both the "
+                    + "changelist and .well-known/first-party-set.json "
+                    + "file: " + str(sorted(field_sym_difference)))
```
cfredric, line 391: Looks like this code is repeated for the associated sites, service sites, and then for each list of ccTLD aliases. Consider factoring out a helper function (or method on the FpsSet class) to construct the error messages. E.g. something like:

```python
def compute_error_from_lists(subset_name, list1, list2):
    if list1 == list2:
        return []
    diff = set(list1) ^ set(list2)   # could be improved if we choose.
    return [f"The following member(s) of {subset_name} were not present in both the PR and the {WELL_KNOWN} file: {str(sorted(diff))}"]
```

Then use it like:

```python
self.error_list.extend(compute_error_from_lists("associatedSites", well_known_set.associated_sites, curr_fps_set.associated_sites))
self.error_list.extend(compute_error_from_lists("serviceSites", well_known_set.service_sites, curr_fps_set.service_sites))
# etc. for ccTLD aliases
```

FpsCheck.py
```diff
@@ -366,33 +364,48 @@ def find_invalid_well_known(self, check_sets):
...
+                if well_known_set.ccTLDs != curr_fps_set.ccTLDs:
+                    field_sym_difference = set(
+                        well_known_set.ccTLDS.keys()) ^ set(
+                            curr_fps_set.ccTLDs.keys())
+                    self.error_list.append("The following aliased ccTLD sites"
+                    + "were not present in both the changelist and " 
+                    + ".well-known/first-party-set.json file: "
+                    + str(sorted(field_sym_difference)))
+                    for aliased_site in curr_fps_set.ccTLDs & well_known_set.ccTLDs:
```
cfredric, line 400: Does this work? AFAIK the `ccTLDs` field is a dict and dicts don't have a `&` operator.

Aside from that - we want to iterate over the union (not the intersection) of the sets of keys, right?

FpsCheck.py
```diff
@@ -366,33 +364,48 @@ def find_invalid_well_known(self, check_sets):
...
+                        field_sym_difference = set(
+                        well_known_set.ccTLDS[aliased_site]) ^ set(
+                            curr_fps_set.ccTLDs[aliased_site])
+                        self.error_list.append("The following ccTLD aliases"
+                        + "were not present in both the changelist and "
```
cfredric, line 405:
```suggestion
                        + "were not present in both the PR and " 
```

"changelist" isn't a term commonly used on GitHub; we should stick to "PR" instead. Same comment elsewhere.

tests/fps_tests.py
```diff
@@ -1336,6 +1343,23 @@ def test_unchecked_field(self, mock_open_and_load_json):
...
+                    "primary": "https://primary5.com",
+                }
+            ]
+        }
+        fp = FpsCheck(fps_sites=json_dict,
+                     etlds=None,
+                     icanns=set())
+        loaded_sets = fp.load_sets()
+        fp.find_invalid_well_known(loaded_sets)
+        self.assertEqual(sorted(fp.error_list), [])
```
cfredric, line 1363: Please add a blank line after this function.

FpsCheck.py
```diff
@@ -366,33 +364,48 @@ def find_invalid_well_known(self, check_sets):
...
+                well_known_set = FpsSet(
+                    json_schema.get('ccTLDs'), 
+                    json_schema.get('primary'), 
+                    json_schema.get('associatedSites'), 
+                    json_schema.get('serviceSites'))
+                if well_known_set.primary != curr_fps_set.primary:
+                    field_sym_difference = [well_known_set.primary, curr_fps_set.primary]
+                    self.error_list.append("The following primary was not "
+                    + "present in both the changelist and "
+                    + ".well-known/first-party-set.json file: "
```
cfredric, line 376:
```suggestion
                    + f"{WELL_KNOWN} file: "
```

Same suggestion in the other places that duplicate the well-known filepath.

FpsCheck.py
sjledoux: I believe they do have the & operator ([an example](https://www.geeksforgeeks.org/python-intersect-two-dictionaries-through-keys/#:~:text=Method%20%232%3A%20Using%20%26%20operator)). I did want to use the intersection here, idea being that we don't want to have a key error when trying to access any site that's aliased on only the well-known or the PR submission. Any `{aliased_site: alias}` pair in which the `aliased_site` is not a key in both the well-known ccTLDs and the submission ccTLDs is automatically considered to be a well-known error without looking at their values, which is taken care of in the [previous step](https://github.com/GoogleChrome/first-party-sets/blob/6c40c62c4a87cc88a02d1b770ab481ead2f39305/FpsCheck.py#L393C21-L395C56). Perhaps the error message there could be altered to include both the sites and their aliases?

FpsCheck.py
sjledoux: Right, I see the issue there. I agree with the thought of printing out the lists to better communicate the issue to the submitter. However, would we just want to do this for the associatedSites field since that is the only one in which the order will have an impact? Or should we do it for the other fields as well?

FpsCheck.py
cfredric: That example uses the `&` operator on `ini_dict1.items()` and `ini_dict2.items()`, not on the dictionaries themselves. There is no `&` operator for dicts:
```python
>>> import sys
>>> sys.version
'3.11.5 (main, Aug 25 2023, 07:43:52) [GCC 12.2.0]'
>>> d = {'x': 'y'}
>>> d & d
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for &: 'dict' and 'dict'
>>> d.items() & d.items()
{('x', 'y')}
```

(The supported operators for dicts are listed below [here](https://docs.python.org/3/library/stdtypes.html#dict.values).)

> I did want to use the intersection here ...

Aha, I see. In that case I think it would be simpler to loop over the union of the keys, and use `aliased_site in dict` or `dict.get(aliased_site, [])` instead of `dict[aliased_site]` to ensure that we don't get a KeyError? That way we could handle all of the aliased sites the same way, instead of handling them differently depending on the shape of the two dicts.

FpsCheck.py
cfredric: I would enforce the ordering for all subsets (+ ccTLDs). I don't see a reason not to, and it's much easier to be strict *now* than to discover we need to be strict later and try to migrate.

FpsCheck.py
sjledoux: Ah, sorry, you're right. I had gotten it mixed up with the `|` operator. 
>> we could handle all of the aliased sites the same way, instead of handling them differently depending on the shape of the two dicts

I'm not sure I fully understand how this would look - if a key is present in one dict and not the other, we would have to treat the process a bit differently regardless as there wouldn't be two values to take the diff of. I suppose incorporating the order matters approach here would alleviate that problem a bit, but the results could overall be less useful. We can talk about this offline to clarify a bit.

FpsCheck.py
```diff
@@ -339,7 +339,16 @@ def check_list_sites(self, primary, site_list):
                 self.error_list.append(
                     "Experienced an error when trying to access " + url + "; "
                     + "error was: " + str(inst))
-                
+    
+    def check_well_known_list(self, field, list1, list2):
+        """Checks that 2 lists for a given field match each other"""
```
cfredric, line 344: Same something about the return value here? E.g. "Returns a list containing all errors found (as strings)."

FpsCheck.py
```diff
@@ -365,31 +372,44 @@ def find_invalid_well_known(self, check_sets):
...
+                service_diff = self.check_well_known_list(
+                    "serviceSites",
+                    curr_fps_set.service_sites, 
+                    well_known_set.service_sites)
+                self.error_list.extend(service_diff)
+                aliased_site_diff = self.check_well_known_list(
+                    "ccTLD key list", 
+                    curr_fps_set.ccTLDs.keys(), 
+                    well_known_set.ccTLDs.keys())
+                self.error_list.extend(aliased_site_diff)
```
cfredric, line 400: I was thinking this info would be conveyed by the loop immediately below this, and we didn't have to check the keys separately.

FpsCheck.py
```diff
@@ -342,6 +340,15 @@ def check_list_sites(self, primary, site_list):
...
+        """Checks that 2 lists for a given field match each other"""
+        if list1 == list2:
+            return []
+        diff = sorted(set(list1) ^ set(list2))
+        return ["Experienced a mismatch between the PR submission and the " 
+        + WELL_KNOWN + " file:\n\t" + field + " was " + str(list1) + " in the PR, and "
+        + str(list2) + " in the well-known.\n\tDiff was: " + str(diff) + "."]
+
     def find_invalid_well_known(self, check_sets):
         """Checks for and validates well-known pages for FPS sets
```
cfredric, line 353: Q: if the checks of a .well-known file fail, do the error messages make it clear to the author _which_ .well-known file they have to fix?

FpsCheck.py
```diff
@@ -329,7 +329,7 @@ def check_list_sites(self, primary, site_list):
                 if 'primary' not in json_schema.keys():
                     self.error_list.append(
                         "The listed associated site site did not have primary"
-                        + " as a key in its .well-known/first-party-set.json "
+                        + " as a key in its " + WELL_KNOWN
```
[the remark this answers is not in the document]
sjledoux: I can agree that the f-string formatting would be better, but I don't want to just change it here and have our string formatting be inconsistent across the repo. I think I would rather make a single change later to address the string formatting everywhere.

FpsCheck.py
cfredric: Right, I'm not saying there's a malignant bug here - just a benign one that happens to not cause problems right now :)

FpsCheck.py
sjledoux: Errors messages for the non-primaries mention the site name, but for the primary they do not.

cfredric: Ack. Doesn't need to be part of this PR, but I think it'd be good if we included the info when checking the primary .well-known as well, so that it's very clear where the errors came from.

cfredric (approved):

FpsCheck.py
iambary44: We have to understand where th error is coming from .it's a big frustration

PR #82, opened 2023-09-13
Refactor some loops to be more idiomatic

cfredric:
This mainly uses different python APIs and takes advantage of the changes to the FpsSet class in 9309ad4.

It does change some error messages to be more accurate, and fixes a bug that could have caused the same error message to be reported multiple times.

### DIFF
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -161,7 +161,7 @@ class FpsCheck:
             else:
                 site_list.update(fps.service_sites)
             # Check the ccTLDs
-            for aliased_site, aliases in fps.ccTLDs.items():
+            for aliases in fps.ccTLDs.values():
                 alias_overlap = set(aliases) & site_list
                 if alias_overlap:
                     self.error_list.append(
--- a/FpsCheck.py
+++ b/FpsCheck.py
@@ -136,8 +136,7 @@ class FpsCheck:
             None
         """
         site_list = set()
-        for primary in check_sets.keys():
-            fps = check_sets[primary]
+        for primary, fps in check_sets.items():
             # Check the primary
             if primary in site_list:
                 self.error_list.append(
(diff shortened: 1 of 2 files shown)


FpsCheck.py
```diff
@@ -469,32 +454,29 @@ def find_invalid_alias_eSLDs(self, check_sets):
             None
         """
         for primary, curr_set in check_sets.items():
-            if not curr_set.ccTLDs:
-                continue
-            for aliased_site in curr_set.ccTLDs:
+            for alias, aliased_sites in curr_set.ccTLDs.items():
```
sjledoux, line 457: I think `alias` and `aliased_site` should be switched here, and in the uses below. Otherwise they're inconsistent with our other references to the field, and with the error messages below (e.g. "The aliased site `alias`"). In general, the key in the pair is the site being aliased, and the value is a list of aliases for that site so I believe `aliased_site`: [aliases] work better as the variable names.

cfredric: Agreed about the mismatch - I was trying to follow the existing usage on lines 208-218 of the original. I've modified all the usages to use the more accurate naming, and fixed some of the error messages to be more accurate.

Also discovered that we were calling `self.check_list_sites` in a loop, once for each aliased site in the set, when we only really have to call it once.

FpsCheck.py
```diff
@@ -469,32 +452,29 @@ def find_invalid_alias_eSLDs(self, check_sets):
             None
         """
         for primary, curr_set in check_sets.items():
-            if not curr_set.ccTLDs:
-                continue
-            for aliased_site in curr_set.ccTLDs:
+            for alias, aliases in curr_set.ccTLDs.items():
```
sjledoux, line 455: I think this should also be `aliased_site, aliases`

cfredric: Done.

sjledoux (approved):
LGTM

PR #84, opened 2023-09-13
Fix path in unit tests to work more generally

sjledoux:
Currently the argument passed to `Sys.path.append()` in `fps_tests.py` may make it difficult for developers to test locally. This PR will make it easier to for developers to run the unit tests on their local repo regardless of what they title their local directory.

### DIFF
--- a/tests/fps_tests.py
+++ b/tests/fps_tests.py
@@ -5,7 +5,7 @@ from publicsuffix2 import PublicSuffixList
 from unittest import mock
 from requests import structures
 
-sys.path.append('../first-party-sets')
+sys.path.append('tests/..')
 from FpsSet import FpsSet
 from FpsCheck import FpsCheck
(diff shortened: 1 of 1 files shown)


tests/fps_tests.py
```diff
@@ -5,7 +5,7 @@
 from unittest import mock
 from requests import structures
 
-sys.path.append('../first-party-sets')
+sys.path.append('tests/..')
```
cfredric, line 8:
```suggestion
sys.path.append('.')
```

This is just a fancy way of saying "the current directory", right?

sjledoux: I guess it is, I had been thinking that the interpreter might initially be reading from the tests directory. It seems silly that it's necessary to append the current directory so that the interpreter can read files from it. I would think that would be the default behavior.

cfredric (approved):

PR #85, opened 2023-09-14
change well-known check to look at rws endpoint

sjledoux:
Now that First Party Sets have [officially be renamed as Related Website Sets](https://developer.chrome.com/blog/related-website-sets/), many of the checks and variable names in the repository should be changed to reflect this, including the well-known check. This PR changes all references to `/.well-known/first-party-set.json` to`/.well-known/related-website-set.json`, primarily affecting `find_invalid_well_known`, `find_invalid_removal`, and their related tests. 

This will change will only affect new/ongoing submissions, older RWS (FPS) entries will not cause new submissions to fail. If an owner of an older RWS entry (meaning one merged before this change) wishes to edit their set, all well-knowns in their set must be migrated to the `/.well-known/related-website-set.json` endpoint in order to pass.


tests/fps_tests.py
```diff
@@ -10,6 +10,8 @@
 from FpsCheck import FpsCheck
 from check_sites import find_diff_sets
 
+WELL_KNOWN = "/.well-known/related-website-set.json"
+
```
cfredric, line 14: Can you rely on the constant directly?

```suggestion
from FpsCheck import FpsCheck, WELL_KNOWN
from check_sites import find_diff_sets

```

FPS-Submission_Guidelines.md
```diff
@@ -138,8 +138,8 @@ Upon submission of a PR, a series of technical checks will run on GitHub to veri
 		<li>Each domain must satisfy the /.well-known/ metadata requirement:</li>
 		<ul>
 <li>The /.well-known/ metadata requirement demonstrates that the submitter has administrative access to the domains present in the set, since administrative access is required to modify the /.well-known/ file. This will help prevent unauthorized actors from adding domains to a set. </li>
```
cfredric, line 140: Should we add a note somewhere about how we'll handle the legacy `.well-known/first-party-set` files going forward?

cfredric (approved):
Approving to unblock, left some comments that can be addressed now or in a followup.

tests/fps_tests.py
sjledoux: Done

FPS-Submission_Guidelines.md
sjledoux: Added a line in the Submission Guideline where it discusses hosting the well-known.

PR #86, opened 2023-09-15
Use fstrings instead of string concatenation for error strings

cfredric:
This makes the error-constructing code a bit easier to read at a glance.


sjledoux (approved):
Looks good from what I can tell. I assume you ran the tests locally, and if they all pass then please merge.

cfredric:
Yep, I did, and the workflow ran here as well. Merging!

PR #96, opened 2023-10-11
Add testing workaround in Getting Started Guide

sjledoux:
Because of changes to the Submission Checks, running `check_sites.py` locally will cause a number of failures to appear related to older submissions that are not compliant with the changes. In order to make the testing process smoother for prospective submitters, this PR adds an explanation of a workaround that can be used to run the checks locally without having to see failures caused by older submissions.

### DIFF
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -19,9 +19,16 @@ To create or revise a submission, you will first need to create a local copy of 
 
 # Testing your Submission Locally #
 
-Once you've made your changes to your local branch, you can open a terminal and run the command "python3 check_sites.py". 
-When this command has finished you will either see "success" meaning your submission passed all of the checks, or you will see a 
-list of failed checks. 
+Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py`. 
+Currently, this will produce a number of check failures due to older submissions in the set failing recent changes to checks. If no members of your set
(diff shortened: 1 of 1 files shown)


cfredric:
The instructions seem fine, but I wonder if we want to adjust the script so that by default, people won't see the errors from the legacy sets if they're running locally. The local runs where those errors are noise will vastly outnumber the local runs where those errors would be important, so I think we ought to optimize for the common case (possibly in a followup PR). WDYT?

cfredric (approved):

sjledoux:
> The instructions seem fine, but I wonder if we want to adjust the script so that by default, people won't see the errors from the legacy sets if they're running locally. 

That could be a possible future action, but not sure what the best way to do that would be? I would think that we'd need to create a separate list that tracks the legacy sets, but that would need to be updated by us as legacy members come and go and would require some reworking of the command-line flow.

I'll merge this for now, and maybe we can create an issue to track possible changes to the flow?

Watza42:
> Because of changes to the Submission Checks, running `check_sites.py` locally will cause a number of failures to appear related to older submissions that are not compliant with the changes. In order to make the testing process smoother for prospective submitters, this PR adds an explanation of a workaround that can be used to run the checks locally without having to see failures caused by older submissions.

sithu864:
https://github.com/GoogleChrome/related-website-sets/pull/96#issuecomment-1758365550

PR #101, opened 2023-10-18

jnm-jai:


Dsbryant (commented):
related_website_sets.JSON

sjledoux:
Hi, thank you for your submission. To clarify some of the failures - your sites are failing the eTLD+1 test because the URLs provided in the associatedSites field are technically all subdomains of eTLD+1s. For most of these, if you were to drop the "www." subdomain from the urls this would stop the cases from failing. This is true for all of the associatedSites except for `https://english.jagran.com`; however, its inclusion is unnecessary since it is a subdomain of `https://jagran.com`. Subdomains are able to get cookie access without being members of a RWS, so you can remove `https://english.jagran.com` entirely from this submission without losing functionality, and you will need to in order to pass the eTLD+1 check.

Finally, the failure in the well-known check is also caused by `https://english.jagran.com`. On `https://jagran.com/.well-known/related-website-set.json` the associated site is listed as "https://english.jagran.com/" with an additional `/` at the end compared to how the site is listed here in the submission. We do require that the text is identical for each site in the submission and the primary's well-known page, so the extra "/" is the source of that failure. Fortunately, since you will already need to remove `https://english.jagran.com` due to the reasons outlined for the previous failure case, this will be easily solved as long as you remove it both from the submission and the well-known page.

jnm-jai:
closing the extra PRs

PR #109, opened 2023-10-24
Add clarification in submission guidelines for subdomains

sjledoux:


RWS-Submission_Guidelines.md
```diff
@@ -134,6 +134,7 @@ Upon submission of a PR, a series of technical checks will run on GitHub to veri
 	<ul>
 <li>Each domain must be prefixed by the https:// scheme. Sets may only include domains served over secure (https://) schemes. </li>
 <li>Each domain must be a <a href="https://github.com/publicsuffix/list/wiki/Format#:~:text=The%20registered%20or%20registrable%20domain%20is%20the%20public%20suffix%20plus%20one%20additional%20label.">registrable domain</a> (i.e., eTLD+1 using a snapshot (refreshed every 6 months) of the <a href="https://publicsuffix.org/">Public Suffix List (PSL)</a> to determine eTLD) at the time of submission. </li>
+<li>Note that subdomains (websites that share the same root structure) are not considered “third-party” by the browser, and do not need to be specified in a set to have access to cookies. For example, shop.example.com and pay.example.com share example.com as the root, and do not need to be in a set to have cookie access. Given that the root example.com is an eTLD+1, if subdomains (shop.example.com, pay.example.com) are added to a set, the eTLD+1 technical checks will fail.</li>
```
cfredric, line 137: Is `root structure` a technical term? Should we say "registrable domain" and link to a definition instead?

cfredric, line 137: Rather than talking about "what the browser considers third-party" (which varies based on the subject [1]), let's just say that Related Websites Sets entries are defined to be registrable domains, rather than origins. We can link to the definitions in the RWS spec if we want to.

[1]: E.g. cookies use "registrable domain" as the boundary, while most other things use "origin" per the [Same Origin Policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).

cfredric, line 137: Should we also consider enhancing the check, so that it detects when something is a subdomain and says something like "`www.foo.com` is not an eTLD+1; it is a subdomain of the `foo.com` eTLD+1." ?

sjledoux: I've been told `root structure` is a phrase used in some of the communications we've had with domain owners, especially in the context of subdomains so I think we wanted to keep the consistency here to try to reduce confusion for those owners.

cfredric: I haven't seen that terminology, but I'll take your word for it.

sjledoux: That does sound reasonable, will add as a future task.

sjledoux: Part of our reason for adding this clarification is that we believe domain owners are confused about the boundaries for third-party cookies. This confusion is leading them to think that they need to add subdomains that are not eTLD+1s into their submissions, thus failing our checks. We wanted to add this note to make it clear that inclusion of such sites is not necessary, which hopefully reduces cases of these failed checks.

cfredric: If we want to explain how the browser imposes boundaries on cookies, then I think we really ought to make the prose specific to *cookies*. Other things (e.g. `localStorage`) in the browser use a different boundary, and it feels misleading to say that the browser's definition of "third-party" is by site.

I would also link to https://web.dev/articles/same-site-same-origin to explain the differences between site/origin/eTLD+1.

We could also clearly state that while cookies are scoped to eTLD+1 in browsers, RWS requires a secure scheme, so in practice RWS requires "sites" that use the `https://` scheme. So, PRs should be written in terms of HTTPS sites.

sjledoux: > We could also clearly state that while cookies are scoped to eTLD+1 in browsers, RWS requires a secure scheme, so in practice RWS requires "sites" that use the `https://` scheme

Are you referring to how if origin A and origin B differ only in scheme, they are considered cross-site? In that regard, don't we essentially have this issue covered by the previous requirement: _Each domain must be prefixed by the https:// scheme_?

cfredric: I'm referring to the mismatch between how browsers draw boundaries for cookies (i.e. by eTLD+1), vs the boundary we expect for RWS (i.e. by site, and specifically only sites that use https:// as the scheme). If we're going to the trouble of explaining origin/eTLD+1/site and the fact that cookies use a coarser boundary than "origin", the natural next question is why RWS doesn't use the same boundary that cookies do.

The answer to that is that RWS uses "eTLD+1 with a secure transport scheme" as the boundary (since `SameSite=None` requires `Secure` which requires a secure transport scheme anyway), and it turns out there's no difference between that and "site with a secure scheme", since a site is just an eTLD+1 and a scheme, by definition.

sjledoux: Rewrote the note to try to incorporate this information. Let me know what you think.

RWS-Submission_Guidelines.md
```diff
@@ -131,16 +131,18 @@ It is important that users' interests are protected from invalid submissions, an
...
-			<li>Example for  primary.com/.well-known/related-website-set.json:</li>
-		</ul></ul>
+<li>Note that RWS uses the qualification of "eTLD+1 with a secure transport scheme" to determine the boundaries of an entry, meaning that `https://example.com` and `https://a.example.com` should not both exist on `related_website_sets.JSON`as they share the same eTLD+1: `example.com`. However, this should not be an issue for the owner of `https://example.com`, as its and `https://a.example.com`'s cookies are not considered third party to each other. Cookies are bounded on same-site, and since `https://a.example.com` is a subdomain of `https://example.com`, it's cookies are not considered cross-site. To learn more about what is considered cross-site vs same-site, [please read this article.](https://web.dev/articles/same-site-same-origin).</li>
```
cfredric, line 137: Pedantic nits:

```suggestion
<li>Note that RWS uses the qualification of "eTLD+1 with a secure transport scheme" to determine the boundaries of an entry, meaning that `https://example.com` and `https://a.example.com` should not both exist on `related_website_sets.JSON` as they share the same eTLD+1: `example.com`. However, this should not be an issue for the owner of `https://example.com`, as its and `https://a.example.com`'s cookies are not considered third-party to each other. Cookies are bounded on same-site, and since `https://a.example.com` is a subdomain of `https://example.com`, its cookies are not considered cross-site. To learn more about what is considered cross-site vs same-site, [please read this article.](https://web.dev/articles/same-site-same-origin).</li>
```

cfredric, line 137:
```suggestion
<li>Note that RWS uses the qualification of "eTLD+1 with a secure transport scheme" to determine the boundaries of an entry, meaning that `https://example.com` and `https://a.example.com` should not both exist on `related_website_sets.JSON`as they share the same eTLD+1: `example.com`. However, this should not be an issue for the owner of `https://example.com`, as its and `https://a.example.com`'s cookies are not considered third party to each other. Cookies are bounded by site, and since `https://a.example.com` is a subdomain of `https://example.com`, both origins have access to the same cookies (modulo the `Domain` attribute) since they are the same site. To learn more about what is considered cross-site vs same-site, [please read this article.](https://web.dev/articles/same-site-same-origin).</li>
```

WDYT about this re-word of one phrase? The existing phrasing feels slightly awkward to me.

cfredric (commented):
This is looking good - I spotted some minor nits and suggested a way to rephrase on part, but overall I'm a fan.

RWS-Submission_Guidelines.md
sjledoux: Applied the suggested changes. Thank you!

RWS-Submission_Guidelines.md
```diff
@@ -131,16 +131,18 @@ It is important that users' interests are protected from invalid submissions, an
...
-			<li>Example for  primary.com/.well-known/related-website-set.json:</li>
-		</ul></ul>
+<li>Note that RWS uses the qualification of "eTLD+1 with a secure transport scheme" to determine the boundaries of an entry, meaning that `https://example.com` and `https://a.example.com` should not both exist on `related_website_sets.JSON`as they share the same eTLD+1: `example.com`. However, this should not be an issue for the owner of `https://example.com`, as its and `https://a.example.com`'s cookies are not considered third-party to each other. Cookies are bounded by site, and since `https://a.example.com` is a subdomain of `https://example.com`, both origins have access to the same cookies (modulo the `Domain` attribute) since they are the same site. To learn more about what is considered cross-site vs same-site, [please read this article.](https://web.dev/articles/same-site-same-origin).</li>
```
cfredric, line 137: Final nit:

```suggestion
<li>Note that RWS uses the qualification of "eTLD+1 with a secure transport scheme" to determine the boundaries of an entry, meaning that `https://example.com` and `https://a.example.com` should not both exist on `related_website_sets.JSON` as they share the same eTLD+1: `example.com`. However, this should not be an issue for the owner of `https://example.com`, as its and `https://a.example.com`'s cookies are not considered third-party to each other. Cookies are bounded by site, and since `https://a.example.com` is a subdomain of `https://example.com`, both origins have access to the same cookies (modulo the `Domain` attribute) since they are the same site. To learn more about what is considered cross-site vs same-site, [please read this article.](https://web.dev/articles/same-site-same-origin).</li>
```

cfredric (approved):
LGTM with one last nit, thanks!

PR #124, opened 2023-11-15
Add unchecked fields to associated .well-known test

cfredric:
This adds a unit test proof that it's fine to include non-`"primary"` fields in the .well-known files for non-primary sites, and even include unsupported fields, without causing a validation error.


sjledoux (approved):
LGTM.

PR #136, opened 2023-11-27
Wirtualna Polska Media related set submission

wojciech-bialy-wpm:


sjledoux:
We noticed formatting edits to other sets. Since edits to sets not maintained by the developer/organization should generally be avoided, we will look into automated formatting improvements.

sjledoux (approved):

PR #150, opened 2023-12-01
Add scheme requirement

sjledoux:
In order to progress with registering the RWS well-known URI, we should specify what URI scheme we require, [as was suggested on the issue we filed](https://github.com/protocol-registries/well-known-uris/issues/40#issuecomment-1835374194).


cfredric (approved):

0192897473:
#150

PR #162, opened 2023-12-12
Fix local testing workflow

sjledoux:
Currently, if users would like to locally test their changes to related_website_sets.JSON, they have to use [an awkward workaround](https://github.com/GoogleChrome/related-website-sets/blob/main/Getting-Started.md#step-4-testing-your-rws-locally). This change makes the workflow much simpler for users by adding an optional command line argument they can pass to check_sites.py when running it locally.

### DIFF
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -41,12 +41,20 @@ Step 4: Testing your RWS locally
 
 Once you've made your changes to your local branch, you can open a terminal and run the command 
 
-`python3 check_sites.py --testing_sets=<your primary site>`
+`python3 check_sites.py --primaries=<your primary site>`
 
-You will get the results of any failed tests in the terminal. Otherwise, you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. If you would like to test multiple Related Website Sets at once, you can run the command above with multiple primaries in a comma separated list, like so:
+You will get the results of any failed tests in the terminal. Otherwise, you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. If you would like to test multiple Related Website Sets at once, you can run the command above with multiple primaries in a comma separated list. 
+
+For example:
+
+`python3 check_sites.py --primaries=https://foo.example`
+
+This would give the results of the checks for a set with `https://foo.example` as its primary.
 
 
-`python3 check_sites.py --testing_sets=<primary 1>,<primary 2>,<etc.>`
+`python3 check_sites.py --primaries=https://foo.example,https://bar.example`
+
+This would give the results for both the set with `https://foo.example` as its primary, and for the set with `https://bar.example` as its primary. 
 
 Step 5: Submitting your RWS
 ---------------------------
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -39,7 +39,14 @@ Step 3: Ensuring your RWS meets the technical requirements
 Step 4: Testing your RWS locally
 --------------------------------
 
-Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py --single_set=<your primary site>`. You will get the restults of any failed tests on the command line. otherwise you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. You can also run the command `python3 check_sites.py`, without the `single_set` argument if you would like, but this will run the checks on the entire document, meaning you may see some failures related to other sets from your own. 
+Once you've made your changes to your local branch, you can open a terminal and run the command 
+
+`python3 check_sites.py --testing_sets=<your primary site>`
+
+You will get the results of any failed tests in the terminal. Otherwise, you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. If you would like to test multiple Related Website Sets at once, you can run the command above with multiple primaries in a comma separated list, like so:
+
+
+`python3 check_sites.py --testing_sets=<primary 1>,<primary 2>,<etc.>`
 
 Step 5: Submitting your RWS
 ---------------------------
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -39,15 +39,7 @@ Step 3: Ensuring your RWS meets the technical requirements
 Step 4: Testing your RWS locally
 --------------------------------
 
-Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py`. Currently, this will produce a number of check failures due to older submissions in the set failing recent changes to checks. If no members of your set are mentioned in the list of failed checks, then your submission should be fine. If you would like a cleaner way of looking for problems in your set, follow these steps:
-
-1.  Run `cp related_website_sets.JSON my_rws.JSON` in your shell
-
-2.  Make your changes to `my_rws.JSON` and make sure `related_website_sets.JSON` is identical to the version in main.
-
-3.  To test your local changes, run `python3 check_sites.py -i my_rws.JSON --with_diff`. When this command finishes running, you will either see "success" (meaning your submission passed all of the checks) or you will see a list of failed checks.
-
-4.  When you are ready to submit, copy your changes from `my_rws.JSON` into `related_website_sets.JSON` and delete `my_rws.JSON`
+Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py --single_set=<your primary site>`. You will get the restults of any failed tests on the command line. otherwise you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. You can also run the command `python3 check_sites.py`, without the `single_set` argument if you would like, but this will run the checks on the entire document, meaning you may see some failures related to other sets from your own. 
 
 Step 5: Submitting your RWS
 ---------------------------
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -41,11 +41,11 @@ Step 4: Testing your RWS locally
 
 Once you've made your changes to your local branch, you can open a terminal and run the command 
 
-`python3 check_sites.py --primaries=<your primary site>`
+`python3 check_sites.py --primaries=https://yourprimary.example`
 
 or, equivalently, run
 
-`python3 check_sites.py -p <your primary site>`
+`python3 check_sites.py -p https://yourprimary.example`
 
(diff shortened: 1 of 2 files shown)


check_sites.py
```diff
@@ -47,16 +47,20 @@ def find_diff_sets(old_sets, new_sets):
 def main():
     args = sys.argv[1:]
     input_file = 'related_website_sets.JSON'
+    testing_sets = []
     input_prefix = ''
     with_diff = False
-    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff"])
+    opts, _ = getopt.getopt(args, "i:", ["data_directory=", "with_diff",
```
cfredric, line 53: Want to allow people to abbreviate this with a short option as well?

Getting-Started.md
```diff
@@ -39,15 +39,14 @@ Step 3: Ensuring your RWS meets the technical requirements
 Step 4: Testing your RWS locally
 --------------------------------
 
-Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py`. Currently, this will produce a number of check failures due to older submissions in the set failing recent changes to checks. If no members of your set are mentioned in the list of failed checks, then your submission should be fine. If you would like a cleaner way of looking for problems in your set, follow these steps:
+Once you've made your changes to your local branch, you can open a terminal and run the command 
 
-1.  Run `cp related_website_sets.JSON my_rws.JSON` in your shell
+`python3 check_sites.py --testing_sets=<your primary site>`
```
cfredric, line 44: A bit of a naming nit, but the "testing" part feels redundant to me, and the "sets" part feels wrong because we're actually asking people to list the primary sites, not sets. So how about:

```suggestion
`python3 check_sites.py --primary=<your primary site>`
```

Getting-Started.md
```diff
@@ -39,15 +39,14 @@ Step 3: Ensuring your RWS meets the technical requirements
...
 
-2.  Make your changes to `my_rws.JSON` and make sure `related_website_sets.JSON` is identical to the version in main.
+You will get the results of any failed tests in the terminal. Otherwise, you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. If you would like to test multiple Related Website Sets at once, you can run the command above with multiple primaries in a comma separated list, like so:
 
-3.  To test your local changes, run `python3 check_sites.py -i my_rws.JSON --with_diff`. When this command finishes running, you will either see "success" (meaning your submission passed all of the checks) or you will see a list of failed checks.
 
-4.  When you are ready to submit, copy your changes from `my_rws.JSON` into `related_website_sets.JSON` and delete `my_rws.JSON`
+`python3 check_sites.py --testing_sets=<primary 1>,<primary 2>,<etc.>`
```
cfredric, line 49:
```suggestion
`python3 check_sites.py --testing_sets="https://foo.example,https://bar.example"`
```

I think it would be helpful to show a more concrete example, so that people don't have to guess whether we require `https://`, whether quotes are necessary, etc.

check_sites.py
```diff
@@ -111,10 +115,17 @@ def main():
...
-        # TODO: add variable and check for subtracted_sets in case of user 
-        # removing old set from the list
     else:
         check_sets = rws_checker.load_sets()
+        if testing_sets:
+            temp_sets = {}
+            for testing_set in testing_sets:
+                if testing_set not in check_sets:
+                    error_texts.append("There was an error loading the set:\n" + testing_set +
+                        " could not be found in related_website_sets.JSON")
```
cfredric, line 125: I think it might be better to print an error and quit entirely here, since the CLI arg was invalid. I'd rather force people to get the CLI arg completely right, than to have them see *some* of the output and (incorrectly) assume everything was fine.

check_sites.py
```diff
@@ -111,10 +115,17 @@ def main():
...
+                else:
+                    temp_sets.update({testing_set: check_sets[testing_set]})
```
cfredric, line 127: FYI, using `.update` is overkill here:

```suggestion
                    temp_sets[testing_set] = check_sets[testing_set]
```

check_sites.py
```diff
@@ -111,10 +115,17 @@ def main():
...
+            check_sets = temp_sets
```
cfredric, line 128: Personal preference, but I'd lean toward writing this in two steps: first validate the arguments, then compute the subset of the `check_sets` dict. E.g.:

```suggestion
        if testing_sets:
            absent_primaries = [p for p in testing_sets if p not in check_sets]
            if absent_primaries:
                # print some error and quit
            check_sets = {p: check_sets[p] for p in testing_sets}
```

check_sites.py
```diff
@@ -47,16 +47,20 @@ def find_diff_sets(old_sets, new_sets):
...
+                                         "testing_sets="])
     for opt, arg in opts:
         if opt == '-i':
             input_file = arg
         if opt == '--data_directory':
             input_prefix = arg
         if opt == '--with_diff':
             with_diff = True
+        if opt == '--testing_sets':
+            testing_sets = arg.split(',')
```
cfredric, line 63:
```suggestion
            testing_sets.extend(arg.split(','))
```

Just in case people supply the arg multiple times.

check_sites.py
sjledoux: As I understand, you're saying in the case where a user enters multiple primaries and one of them is not in the RWS list, we should exit without giving error information about the sets whose primaries they entered correctly? If so, I think I disagree - I'd prefer we give us much error information as possible when they run the command. My guess is that most users are only going to be running the command with a single primary argument, so this would be moot anyway.

cfredric: Agree that it's probably rare that people will supply more than one site.

My fear about showing as much error information as possible in this case is that I think it would be easy to miss this error, in the midst of others. But I guess if there are other errors, then the dev will still know that there's something they need to fix; so if everything else is fine (and there is no other output), then it'll be easy to spot this error, anyway. So I'm ok with making this error non-fatal.

Getting-Started.md
```diff
@@ -39,15 +39,22 @@ Step 3: Ensuring your RWS meets the technical requirements
...
 
-2.  Make your changes to `my_rws.JSON` and make sure `related_website_sets.JSON` is identical to the version in main.
+You will get the results of any failed tests in the terminal. Otherwise, you will see "success" if your changes are passing all of the checks. Make sure that the text of the primary site you're passing into the command line is identical to the primary site you have listed in the related_website_sets.JSON file, or the tests will fail. If you would like to test multiple Related Website Sets at once, you can run the command above with multiple primaries in a comma separated list. 
 
-3.  To test your local changes, run `python3 check_sites.py -i my_rws.JSON --with_diff`. When this command finishes running, you will either see "success" (meaning your submission passed all of the checks) or you will see a list of failed checks.
+For example:
 
-4.  When you are ready to submit, copy your changes from `my_rws.JSON` into `related_website_sets.JSON` and delete `my_rws.JSON`
+`python3 check_sites.py --primaries=https://foo.example`
+
```
cfredric, line 51: We probably should show the short version too, like this I think:

```suggestion
`python3 check_sites.py --primaries=https://foo.example`
Equivalently:
`python3 check_sites.py -p https://foo.example`
```

(Please check that I'm understanding `getopt` correctly :))

check_sites.py
```diff
@@ -111,10 +115,14 @@ def main():
...
-        # TODO: add variable and check for subtracted_sets in case of user 
-        # removing old set from the list
     else:
         check_sets = rws_checker.load_sets()
+        if cli_primaries:
+            absent_primaries = [p for p in cli_primaries if p not in check_sets]
+            for p in absent_primaries:
+                error_texts.append("There was an error loading the set:\n" + 
+                    p + " could not be found in related_website_sets.JSON")  
+            check_sets = {p: check_sets[p] for p in cli_primaries if p not in absent_primaries}
```
cfredric, line 125: If one of the cli_primaries isn't in `absent_primaries`, that means it *is* in `check_sets`, so this feels a bit more direct/clear to me:

```suggestion
            check_sets = {p: check_sets[p] for p in cli_primaries if p in check_sets}
```

Getting-Started.md
```diff
@@ -39,15 +39,22 @@ Step 3: Ensuring your RWS meets the technical requirements
...
+This would give the results of the checks for a set with `https://foo.example` as its primary.
+
+
+`python3 check_sites.py --primaries=https://foo.example,https://bar.example`
+
```
cfredric, line 56: Similar comment as above, using the short version:

```suggestion
`python3 check_sites.py --primaries=https://foo.example,https://bar.example`

Equivalently:
`python3 check_sites.py -p https://foo.example -p https://bar.example`
```

check_sites.py
```diff
@@ -111,10 +115,14 @@ def main():
...
         check_sets, subtracted_sets = find_diff_sets(old_checker.load_sets(), rws_checker.load_sets())
-        # TODO: add variable and check for subtracted_sets in case of user 
-        # removing old set from the list
     else:
         check_sets = rws_checker.load_sets()
+        if cli_primaries:
+            absent_primaries = [p for p in cli_primaries if p not in check_sets]
+            for p in absent_primaries:
+                error_texts.append("There was an error loading the set:\n" + 
+                    p + " could not be found in related_website_sets.JSON")
```
cfredric, line 124: Strictly speaking, `p` might be in related_website_sets.JSON as a non-primary site, right? So how about:

```suggestion
                error_texts.append("There was an error loading the set:\n" + 
                    f"could not find a set with primary site \"{p}\"")
```

sjledoux: Done.

Getting-Started.md
sjledoux: Done.

check_sites.py
sjledoux: Done.

Getting-Started.md
sjledoux: Done.

Getting-Started.md
```diff
@@ -39,15 +39,32 @@ Step 3: Ensuring your RWS meets the technical requirements
 Step 4: Testing your RWS locally
 --------------------------------
 
-Once you've made your changes to your local branch, you can open a terminal and run the command `python3 check_sites.py`. Currently, this will produce a number of check failures due to older submissions in the set failing recent changes to checks. If no members of your set are mentioned in the list of failed checks, then your submission should be fine. If you would like a cleaner way of looking for problems in your set, follow these steps:
+Once you've made your changes to your local branch, you can open a terminal and run the command 
 
-1.  Run `cp related_website_sets.JSON my_rws.JSON` in your shell
+`python3 check_sites.py --primaries=<your primary site>`
```
cfredric, line 44: Personal preference, I would make this example concrete as well, since I expect most people will only read down to here.

check_sites.py
```diff
@@ -111,10 +115,14 @@ def main():
...
         check_sets, subtracted_sets = find_diff_sets(old_checker.load_sets(), rws_checker.load_sets())
-        # TODO: add variable and check for subtracted_sets in case of user 
-        # removing old set from the list
     else:
         check_sets = rws_checker.load_sets()
+        if cli_primaries:
+            absent_primaries = [p for p in cli_primaries if p not in check_sets]
+            for p in absent_primaries:
+                error_texts.append("There was an error loading the set:\n" + 
+                    "could not find set with primary site " + p)
```
cfredric, line 124: `p` is coming from the user, and we're looking for exact equality, so we ought to use quotes so that any whitespace or other mismatches are easier to spot.

```suggestion
                    f"could not find set with primary site \"{p}\"")  
```

Getting-Started.md
sjledoux: Done.

check_sites.py
sjledoux: Done.

cfredric (approved):
LGTM, thanks! (Go ahead and merge when ready)

PR #189, opened 2023-12-28

miguelborgez:
Adds talkdeskqaid.com and trytalkdesk.com as related sites.

### DIFF
--- a/related_website_sets.JSON
+++ b/related_website_sets.JSON
@@ -346,6 +346,14 @@
           "https://journaldunet.fr"
         ]
       }
+    },
+    {
+      "primary": "https://talkdeskqaid.com",
+      "contact": "support@talkdesk.com",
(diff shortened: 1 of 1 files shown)


sjledoux (approved):
LGTM, merging.

cfredric:
Hi @miguelborgez,

It looks like https://trytalkdesk.com/.well-known/related-website-set.json gives a 404 error now. This URL is [expected to serve](https://github.com/GoogleChrome/related-website-sets/blob/main/RWS-Submission_Guidelines.md#set-level-technical-validation) the JSON file for the duration of trytalkdesk.com's inclusion in the set.

Can you please update your site to serve the JSON from this URL again? Otherwise we cannot verify that trytalkdesk.com intends to be included in your RWS, so we will remove it from your set entry.

Issue #246, opened 2024-02-01
find_robots_txt check does not work with redirects or reflect the test description


erikb-stripe:
The `find_robots_txt` check does not work properly with redirects.

The description states
```
        Iterates through all service_sites in each RwsSet provided, and makes
        a get request to site/robots.txt for each. This request should return
        an error 4xx, 5xx, or a timeout error. If it does not, and the page 
        does exist, then it is expected that the site contains a X-Robots-Tag
        in its header. If none of these conditions is met, an error is appended
        to the error list.
```

However the [code](https://github.com/GoogleChrome/related-website-sets/blob/main/RwsCheck.py#L485) only makes a request to the root, it does not first check for a 4xx/5xx against the service domain.

```
r_service = requests.get(service_site, timeout=10)
```

If a redirect ([per the guidelines](https://github.com/GoogleChrome/related-website-sets/blob/main/RWS-Submission_Guidelines.md#:~:text=Must%20have%20a%20homepage%20that%20redirects%20to%20a%20different%20domain%20or%20results%20in%204xx%20(client%20error)%20or%205xx%20(server%20error).)) is in place then the headers being checked are for the destination domain, not the service one.

I'm working to set up for our service domains to meet the requirements but running into issues because of the structure of this test and am waiting to finalize the setup until I know the expectations of the check.

erikb-stripe:
I found https://github.com/GoogleChrome/related-website-sets/pull/28 which caused this breakage and made the comment out-of-date.

I think the solution here is to either allow robots.txt files that have noindex / none rules OR require check that the root of the domain returns a `X-Robots-Tag` header.

PR #247, opened 2024-02-02
Prevents redirects in the find_robots_txt check

erikb-stripe:
Fixes https://github.com/GoogleChrome/related-website-sets/issues/246 by explicitly disabling redirects on the `find_robots_txt` check. Adds asserts that this is the case in the tests.

This allows for the case where a service domain redirects to another domain that does not have the `X-Robots-Tag`.

This shouldn't change the behavior of non-redirecting which must either 4xx or 5xx when the home page is hit.

> Must have a homepage that redirects to a different domain or results in 4xx (client error) or 5xx (server error).

### DIFF
--- a/tests/rws_tests.py
+++ b/tests/rws_tests.py
@@ -921,6 +921,11 @@ def mock_get(*args, **kwargs):
         mgr = MockedGetResponse({}, 200)
         mgr.url = 'https://example.com'
         return mgr
+    elif args[0] == 'https://service7.com':
+        if 'allow_redirects' in kwargs:
+            if kwargs['allow_redirects'] != True:
+                return MockedGetResponse({"X-Robots-Tag":"noindex"}, 200)
+        return MockedGetResponse({"X-Robots-Tag":"foo"}, 200) 
     elif args[0].startswith('https://service'):
         return MockedGetResponse({},200)
     elif args[0] == 'https://primary1.com' + WELL_KNOWN:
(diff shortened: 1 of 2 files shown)


tests/rws_tests.py
```diff
@@ -1062,6 +1065,7 @@ def test_robots_none_tag(self, mock_get):
         loaded_sets = rws_check.load_sets()
         rws_check.find_robots_txt(loaded_sets)
         self.assertEqual(rws_check.error_list, [])
+        mock_get.assert_called_once_with('https://service4.com', timeout=10, allow_redirects=False)
```
sjledoux, line 1069: Instead of inserting the calls to `assert_called_once_with` into the pre-existing robots tests, I would prefer a single test case that would produce a different result if not called with `allow_redirects=False`, such as adding the following case to line 924: 
```
    elif args[0] == 'https://service7.com':
        if 'allow_redirects' in kwargs:
            if kwargs['allow_redirects'] != True:
                return MockedGetResponse({"X-Robots-Tag":"noindex"}, 200)
        return MockedGetResponse({"X-Robots-Tag":"foo"}, 200) 
```
And inserting the following test here:

```suggestion
    
    @mock.patch('requests.get', side_effect=mock_get)
    def test_robots_redirects(self, mock_get):
        json_dict = {
            "sets":
            [
                {
                    "primary": "https://primary.com",
                    "serviceSites": ["https://service7.com"]
                }
            ]
        }
        rws_check = RwsCheck(rws_sites=json_dict,
                     etlds=None,
                     icanns=set())
        loaded_sets = rws_check.load_sets()
        rws_check.find_robots_txt(loaded_sets)
        self.assertEqual(rws_check.error_list, [])
        
```

sjledoux (changes requested):
Thank you for finding this bug! I agree with the change, but would prefer a different way of handling the testing in `rws_tests.py`

tests/rws_tests.py
erikb-stripe: Ack. Just made these changes. Thanks for the review.

sjledoux (approved):
LGTM, merging.

PR #248, opened 2024-02-02
Avoid following redirects when checking X-Robots-Tag on service sites

griff-stripe:


sjledoux:
Closing this as the same suggested change was applied in #247.

PR #257, opened 2024-02-06
Fix bug with concatenating JSONDecodeError

sjledoux:
JSONDecodeErrors were not properly handled in the `check_sites.py` script, meaning that the comment-bot wasn't surfacing the error to the submitters. This change fixes the formatting.

### DIFF
--- a/check_sites.py
+++ b/check_sites.py
@@ -68,8 +68,8 @@ def main():
             rws_sites = json.load(f)
         except Exception as inst:
         # If the file cannot be loaded, we will not run any other checks
-            print("There was an error when loading "+ input_file + 
-                  "\nerror was: " + inst)
+            print(f"There was an error when loading {input_file};" 
+                  f"\nerror was:  {inst}")
             return  
      
 


cfredric:
I'm not seeing how this fixes the bug - it looks like it just adds a semicolon to the message we print to stdout. But the "Create comment" step just compares against "success", so IIUC the current message should still be triggering a comment.

Is the current code not WAI because we're matching the wrong kind of error, i.e. we should look for `JSONDecodeError` or `ValueError` instead of `Exception`?

Also, is it possible to test this?

cfredric:
Ah, never mind - I see that the issue was that the code was trying to concat `str + JSONDecodeError`, instead of `str + str(JSONDecodeError)`. Fix LGTM then.

It'd still be nice if this were tested though, so that we can prevent regressions in this code in the future.

sjledoux:
> It'd still be nice if this were tested though, so that we can prevent regressions in this code in the future.

I did test this locally, if that offers some comfort. I do agree that we need some genuine integration tests for the whole workflow process though.

Hamedpahlan (commented):
Revawe

PR #263, opened 2024-02-08
Clarification PR: Tech Checks in Dev Testing

thegreatfatzby:
I'm willing to bet this is the case so am making a PR rather than an issue...if I'm wrong please reject to lmk :)


aselya:
That assumption is not the case; Chrome clients do verify some of the invariants that are also enforced by the workflows in this repository. But in general, we'd prefer to avoid making promises here about what Chrome does or does not validate, since those details may change in the future. As such, we'll reject this PR.

PR #279, opened 2024-02-20
Trigger "RWS submission checks" action daily at 9am EST

aselya:
This commit will cause the "RWS submission checks" to run daily at 9am EST. It utilizes the "schedule" event trigger and cron as described in https://docsgithub.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

### DIFF
--- a/.github/workflows/rws-submissions-checks.yml
+++ b/.github/workflows/rws-submissions-checks.yml
@@ -14,7 +14,7 @@
 name: RWS submission checks
 on: 
   schedule:
-  - cron: '0 14 * * *' #triggers job at 9am EST each day
+  - cron: '0 14 * * 2' # 2pm UTC each Tuesday
   pull_request_target:
 jobs:
   PR-Actions:
--- a/.github/workflows/rws-submissions-checks.yml
+++ b/.github/workflows/rws-submissions-checks.yml
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 name: RWS submission checks
-on: pull_request_target
+on: 
+  schedule:
+  - cron: '0 14 * * *' #triggers job at 9am EST each day
+  pull_request_target:
 jobs:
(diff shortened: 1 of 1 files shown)


aselya:
Limitations:
The trigger time is set as 9am EST but there appears to be [no assurance that the job will run ](https://github.com/orgs/community/discussions/27130) at 9am. If this isn't running by the time pull requests get reviewed the time should be changed to something earlier.

It also will not apply to any pull requests that were created before the change gets merged in so reviewers should still confirm that the checks have run.

.github/workflows/rws-submissions-checks.yml
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 name: RWS submission checks
-on: pull_request_target
+on: 
+  schedule:
+  - cron: '0 14 * * *' #triggers job at 9am EST each day
```
cfredric, line 17: We don't look at the PRs except on Tuesdays, so I don't think we need to have daily jobs. Just once weekly, before we review the PRs would be sufficient.

(I also think it'd be good to be precise about the time/timezone, so that it's clearer what the crontab is supposed to be.)

```suggestion
  - cron: '0 14 * * 2' # 2pm UTC each Tuesday
```

cfredric (approved):
LGTM, go ahead and merge when ready

12shraf (commented):
4244540065385265
12/28
البنك العربي الإسلامي الدولي 
Send me a message on this bank account and your work with me so that the desired amount has been deposited. I will deposit the amount for you twice, with thanks, Sharaf Al-Zoubi.

arichiv:
This automated check has never passed because it runs on the main branch and not pull requests: https://github.com/GoogleChrome/related-website-sets/actions/workflows/rws-submissions-checks.yml?page=2&query=event%3Aschedule

PR #299, opened 2024-03-05
Add entry for CarCostAdvisor

KVermoteEFC:


cfredric (approved):
LGTM

related_website_sets.JSON
```diff
@@ -572,6 +572,19 @@
...
+      "primary": "https://carcostadvisor.com",
+      "associatedSites": [],
+      "serviceSites": [],
+      "rationaleBySite": {},
+      "ccTLDs": {
+        "https://carcostadvisor.com": [
+          "https://carcostadvisor.be",
+          "https://carcostadvisor.fr"
+        ]
+      }
```
Cre-Dy, line 586: These seem to be connected to non authorized individuals can you please trace and forward to proper authorities.

KVermoteEFC: Hi,
I'm not quite sure what you mean as I'm the administrator of these websites (which of course enabled me to add the .well-known/related-website-set.json files in the first place).

Maybe I should point out though that we don't actually use the apex domains carcostadvisor.be and carcostadvisor.fr - these simply redirect the user to carcostadvisor.com, which contains our general welcome page.
Our tenant users always use their designated subdomain, like bmw.carcostadvisor.be or renault.carcostadvisor.fr, etc.

I used to do the redirecting on the DNS-level, but as the RWS workflow expects the https://carcostadvisor.be/.well-known/related-website-set.json URL to work, I had to create 2 static websites for the apex domains, which contain only the "related-website-set.json" and an "index.html" that redirects the user to the carcostadvisor.com website as before.

Does this satisfy your comment?

KVermoteEFC: FYI, the static websites are hosted in Azure Storage.

hamadilubinga:
Je demande l'aide de traducteurs pour la suite et des conseils pour progresser

Labestia (commented):
[Gmail - GoogleDriveSetup.exe.pdf](https://github.com/GoogleChrome/related-website-sets/files/14894691/Gmail.-.GoogleDriveSetup.exe.pdf)

PR #322, opened 2024-03-20
Create VIE-Related-Websites

ARSGeek:
This is new territory for me. So I'm flying blind a bit. A client's website has an 3rd-party i-frame form on it that is being blocked by Google Chrome on iOS only. My research brought me to the Google Developers Related Website Set. Which brought me here. I think I'm doing this the right way. I just want to clear up the issue with the client's website.

### DIFF
--- a/VIE-Related-Websites
+++ b/VIE-Related-Websites
@@ -1,0 +1,21 @@
+{
+  "sets": [
+    {
+      "contact": "adam@arsdigitalmarketing.com",
+      "primary": "https://viemagazine.com",
+
+      "associatedSites": ["https://www.google-analytics.com", "https://theideaboutique.maghub.com", "https://theideaboutique.adorbit.com", "https://www.google.com/recaptcha"],
+
+      "serviceSites": ["https://theideaboutique.com"],
+
+      "rationaleBySite": {
+        "https://www.google-analytics.com": "Google analytics is needed for obvious reasons, tracking website traffic, interactions, etc.",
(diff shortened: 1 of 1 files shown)


cfredric:
Hi @ARSGeek, the biggest issue with this PR is that it creates a brand new file, instead of modifying the existing `related_website_sets.JSON` file. Please follow the instructions [here](https://github.com/GoogleChrome/related-website-sets/blob/main/Getting-Started.md#step-2-creating-or-updating-your-rws-submission) in order to get your set ready for submission.

However, based on your description, I'm not sure that RWS is a viable solution to your problem.

Firstly, RWS is not supported in Chrome on iOS (for technical reasons). So, even if this set were submitted as-is, it wouldn't change the behavior you're seeing.

Secondly, your client will need to do some work to prepare their sites (assuming they control both the top-level site and the iframe form's site) for being in an RWS, namely they need to satisfy the [`.well-known` requirements](https://github.com/GoogleChrome/related-website-sets/blob/main/Getting-Started.md#step-2-creating-or-updating-your-rws-submission). But considering you say the iframe is 3rd-party, it doesn't sound like the client controls both of those sites, ultimately.

ARSGeek:
@cfredric Thank you for your reply. This may not be even the issue since the 3rd-party form works fine on all Desktop versions and well as android. It is only iPhones that don't show the form. I may put this on hold and investigate with the 3rd-party vendor.

cfredric:
It looks like you haven’t updated this PR in a while. Your PR will close in 14 days if there aren’t any updates or comments on this PR.

arichiv:
Due to inactivity, we are closing this PR. To restart the process, please submit a new PR.

Issue #339, opened 2024-04-12
Technical checks should issue a warning for sets with more than 5 associated sites


krgovind:
> I'm a bit surprised about the 5-domain cap + organizations being allowed to submit more than 5 domains without a warning.

Capturing feedback from #285 from @thezedwards that it will be helpful for the technical checks to issue a warning on sets with more than 5 associated sites; since as per [currently documented behavior](https://github.com/GoogleChrome/related-website-sets/blob/main/RWS-Submission_Guidelines.md#browser-behavior), the browser will auto-reject rSA/rSAFor calls beyond the limit. Developers submitting larger sets may be surprised by this.

Redj34:
- https://github.com/GoogleChrome/related-website-sets.git

PR #373, opened 2024-05-07
Add associated warning

sjledoux:
As per #339, the checks should emit a warning when submissions contain more than 5 associated sites. This PR adds the warning to the automated comment left on PRs which informs submitters of success or failure.

### DIFF
--- a/.github/workflows/rws-submissions-checks.yml
+++ b/.github/workflows/rws-submissions-checks.yml
@@ -64,7 +64,7 @@ jobs:
           message-path: "message.txt" 
           refresh-message-position: true
       - name: Fail or Succeed
-        if: steps.read_results.outputs.contents != 'success'
+        if: startsWith(steps.read_results.outputs.contents, 'It appears you have failed')
         uses: actions/github-script@v3
         with:
           script: |
--- a/.github/workflows/rws-submissions-checks.yml
+++ b/.github/workflows/rws-submissions-checks.yml
@@ -46,13 +46,18 @@ jobs:
           CONTENTS: ${{ steps.read_results.outputs.contents }}
         run: echo "$CONTENTS"
       - name: Create Comment
-        if: steps.read_results.outputs.contents != 'success'
+        if: steps.read_results.outputs.contents != 'success' && ${{ !startsWith(steps.read_results.outputs.contents, 'Warning')}}
         run: |
          echo "It appears you have failed some tests. Here are your results:" > message.txt
          cat results.txt >> message.txt
-      - name: Comment if sucess
+      - name: Comment if success
         if: steps.read_results.outputs.contents == 'success'
         run: echo "Looks like you've passed all of the checks!" >> message.txt
+      - name: Comment if warning
+        if: startsWith(steps.read_results.outputs.contents, 'Warning')
+        run: |
+          echo "You have passed all of the checks, but there are one or more warnings associated with your submission." > message.txt
+          cat results.txt >> message.txt
       - name: Write Comment on PR
         uses: mshick/add-pr-comment@v2
         with:   
--- a/RwsCheck.py
+++ b/RwsCheck.py
@@ -46,7 +46,7 @@ class RwsCheck:
         self.etlds = etlds
         self.icanns = icanns
         self.error_list = []
-        self.associated_warning = []
+        self.warning_texts = []
 
     def validate_schema(self, schema_file):
         """Validates the canonical sites list
@@ -185,7 +185,7 @@ class RwsCheck:
         """
         for primary, rws in check_sets.items():
             if len(rws.associated_sites) > ASSOCIATED_LIMIT:
-                self.associated_warning.append(
+                self.warning_texts.append(
                     f"Warning: the set for {primary} contains more than {ASSOCIATED_LIMIT} associated sites."
                 )
 
--- a/RwsCheck.py
+++ b/RwsCheck.py
@@ -20,6 +20,7 @@ from urllib.request import Request
 from publicsuffix2 import PublicSuffixList
 
 WELL_KNOWN = "/.well-known/related-website-set.json"
+ASSOCIATED_LIMIT = 5
 
 class RwsCheck:
 
@@ -45,6 +46,7 @@ class RwsCheck:
         self.etlds = etlds
         self.icanns = icanns
         self.error_list = []
+        self.associated_warning = []
 
     def validate_schema(self, schema_file):
         """Validates the canonical sites list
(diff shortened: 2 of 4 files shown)


.github/workflows/rws-submissions-checks.yml
```diff
@@ -46,20 +46,25 @@ jobs:
           CONTENTS: ${{ steps.read_results.outputs.contents }}
         run: echo "$CONTENTS"
       - name: Create Comment
-        if: steps.read_results.outputs.contents != 'success'
+        if: steps.read_results.outputs.contents != 'success' && ${{ !startsWith(steps.read_results.outputs.contents, 'Warning')}}
```
cfredric, line 49: Does this support emitting a warning alongside errors? What if the warning happens to be the first line that was emitted, but it's followed by errors?

(Maybe we ought to write separate files for warnings/errors.)

.github/workflows/rws-submissions-checks.yml
```diff
@@ -46,20 +46,25 @@ jobs:
...
+          echo "You have passed all of the checks, but there are one or more warnings associated with your submission." > message.txt
+          cat results.txt >> message.txt
       - name: Write Comment on PR
         uses: mshick/add-pr-comment@v2
         with:   
           message-path: "message.txt" 
           refresh-message-position: true
       - name: Fail or Succeed
-        if: steps.read_results.outputs.contents != 'success'
+        if: startsWith(steps.read_results.outputs.contents, 'It appears you have failed')
```
cfredric, line 67: Can all of this be moved into the "Create Comment" step, so that we don't have to duplicate the "It appears you have failed" substring? Duplicating that makes this brittle, it'll break if someone changes that string and doesn't realize there's a dependency here.

RwsCheck.py
```diff
@@ -170,6 +172,23 @@ def check_exclusivity(self, check_sets):
...
 
+    def check_associated_count(self, check_sets):
+        """This method checks for RwsSets that exceed the associated limit
+
+        Creates a warning for each set passed in the check_sets list that has 
+        more associatedSites than the ASSOCIATED_LIMIT
+
+        Args:
+            check_sets: Dict[string, RwsSet]
+        Returns:
```
cfredric, line 183: For future refactoring: IMO it'd be better design to make this a ["pure" function](https://en.wikipedia.org/wiki/Pure_function), i.e. it takes all its inputs as arguments, and returns its result explicitly (rather than just mutating the instance). That makes it easier to test, and easier for a reader to understand how to use the method.

Doing that in this CL would require the caller to know whether the results should be considered warnings or errors, though, which isn't great. What'd be nice is if each check returned a list of errors *and* and list of warnings, and the caller just appended each to its running lists. That's a larger change, so it shouldn't be part of this PR.

check_sites.py
```diff
@@ -145,12 +146,14 @@ def main():
...
             error_texts.append(inst)
     # This message allows us to check the succes of our action
-    if rws_checker.error_list or error_texts:
+    if rws_checker.error_list or error_texts or rws_checker.warning_texts:
         for checker_error in rws_checker.error_list:
             print(checker_error)
         for error_text in error_texts:
             print(error_text)
-    else:
+        for warning in rws_checker.warning_texts:
```
cfredric, line 154: Maybe add a comment here that indicates it's important for the warnings to be *after* errors? Unless you choose to remove that limitation, per my other comment.

check_sites.py
```diff
@@ -145,12 +146,14 @@ def main():
...
+            print(warning)
```
cfredric, line 155: This is getting complicated/repetitive; consider extracting the things we want to print:

```python
outputs = rws_checker.error_list + error_texts + rws_checker.warning_texts
if outputs:
  for output in outputs:
    print(output)
else:
  print("success", end="")
```

Note that this no longer repeats each list twice.

tests/rws_tests.py
```diff
@@ -276,6 +276,84 @@ def test_expected_rationales_case(self):
...
+                    "rationaleBySite": {}
+                }
+            ]
+        }
+        rws_check = RwsCheck(rws_sites=json_dict,
+                      etlds=None,
+                       icanns=set())
+        loaded_sets = rws_check.load_sets()
+        rws_check.check_associated_count(loaded_sets)
+        self.assertEqual(rws_check.associated_warning, [])
```
cfredric, line 296: This assertion is more specific than it needs to be, so it is a [change-detector](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) assertion. (If we add another kind of warning in the future, this assertion may fail even though the condition that it's meant to check -- that the >5 associated sites warning is not present -- is still true.)

Please check for the specific warning that this test cares about, not that there are 0 warnings overall.

tests/rws_tests.py
```diff
@@ -276,6 +276,84 @@ def test_expected_rationales_case(self):
...
+                    "rationaleBySite": {}
+                }
+            ]
+        }
+        rws_check = RwsCheck(rws_sites=json_dict,
+                      etlds=None,
+                       icanns=set())
+        loaded_sets = rws_check.load_sets()
+        rws_check.check_associated_count(loaded_sets)
+        self.assertEqual(rws_check.associated_warning,
```
cfredric, line 319: Please check that the warning we want is *among* the actual warnings; not that it is the *only* warning. (You could do this with `self.assertIn` or `self.assertCountEqual`.)

Same comment for the test case below.

tests/rws_tests.py
cfredric, line 296: Where is `rws_check.associated_warning` defined? It looks like you meant this to be `rws_check.warning_texts` instead, here and below. If so, please investigate why the tests passed despite this typo?

.github/workflows/rws-submissions-checks.yml
An010299: Do you want help me?

An010299 (commented):
What's need?

.github/workflows/rws-submissions-checks.yml
ratchanonsuttawas: > Does this support emitting a warning alongside errors? What if the warning happens to be the first line that was emitted, but it's followed by errors?
> 
> (Maybe we ought to write separate files for warnings/errors.)

ratchanonsuttawas (commented):
Changes

.github/workflows/rws-submissions-checks.yml
ratchanonsuttawas, line 1: Change

RwsCheck.py
ratchanonsuttawas, line 1: Chang

check_sites.py
ratchanonsuttawas, line 1: Change

tests/rws_tests.py
ratchanonsuttawas, line 1: Changes

ratchanonsuttawas (commented):
Summit feedback and approve merging these changes

.github/workflows/rws-submissions-checks.yml
0192897473: bari saya petunjuk. bantu saya dalam proses ini.     semua keputusan di tangan saya.  nama anda akan saya kait dalam keberhasilan saya.

sjledoux:
Closing this as I have not updated it in a while. I will re-open this when I have a new commit.

PR #380, opened 2024-05-11

oliverterrell:
Encrypted JWT tokens are used in the primary serverless architecture for authenticated sessions. I need this value present in requests to my custom server as well for security reasons. This server is needed for server-based use cases that are inaccessible otherwise. This cookie is one of the good guys.


shuranhuang (approved):
LGTM

PR #414, opened 2024-06-09
added domains related to text.com

lukiq:


lukiq:
Close nad get back to it after I will change the main file.

PR #502, opened 2024-08-12
Create generator-generic-ossf-slsa3-publish.yml

iamwahyu17:


CrabbyJim:
Thank you for your response. If possible I would like to recreate the
program I had before with this platform 🤔

PR #530, opened 2024-09-06
Create SECURITY.md

Benz19933:


cfredric:
Closing, since this looks copy/pasted from some template and this repo does not have any versioning scheme (so the `5.1.x`, `4.0.x`, etc.) in the file are meaningless.

PR #671, opened 2024-12-09
Validate JSON formatting

arichiv:
Let's compare the input to [json.dump](https://docs.python.org/3/library/json.html#json.dump) as our source of truth. This should prevent issues with excess whitespace, inconsistent newlines or indentation, and more.

Because the actions are controlled by the master branch, I had to run the checks locally to validate. If they're run with the current JSON changes we do get 'success' in the results.txt file.


check_sites.py
```diff
@@ -70,8 +74,21 @@ def main():
         # If the file cannot be loaded, we will not run any other checks
             print(f"There was an error when loading {input_file};" 
                   f"\nerror was:  {inst}")
-            return  
-     
+            return
+        # Notify of any formatting errrors in the JSON
```
miketaylr, line 78:
```suggestion
        # Notify of any formatting errors in the JSON
```

miketaylr (commented):
Nice. If we do land this, we should probably document expectations somewhere (e.g., RWS-Submission_Guidelines.md)

check_sites.py
```diff
@@ -70,8 +74,21 @@ def main():
...
+        # Notify of any formatting errors in the JSON
+        if with_format:
+            f.seek(0)
+            loaded_file = f.read()
+            # Add final newline by convention
+            formatted_file = json.dumps(rws_sites, indent=2, ensure_ascii=False) + "\n"
+            if loaded_file != formatted_file:
+                diff = difflib.ndiff(loaded_file.splitlines(keepends=True), formatted_file.splitlines(keepends=True))
+                # Only show lines with differences
+                filtered_diff = filter(lambda line: len(line) > 0 and line[0] != ' ', diff)
```
cfredric, line 87: Generator comprehensions are "more pythonic" (https://google.github.io/styleguide/pyguide.html#210-lambda-functions).
```suggestion
                filtered_diff = (line for line in diff if len(line) > 0 and line[0] != ' ')
```

check_sites.py
```diff
@@ -12,12 +12,44 @@
...
 # limitations under the License.
 from RwsCheck import RwsCheck
+import difflib
 import json
 import getopt
 import sys
 import os
 from publicsuffix2 import PublicSuffixList
 
+def load_rws_file_as_json(input_file, with_format):
```
cfredric, line 22: Nit: preexisting problem, but `input_file` is ambiguous - is that the name of the file or the contents of it? IMO we should call this `input_filename` or something.

tests/rws_tests.py
```diff
@@ -9,7 +9,26 @@
...
+        self.assertEqual(load_rws_file_as_json("tests/bad_formatting.JSON", False),
+                         ({"a": "foo", "b": "bar"}, None))
+        self.assertEqual(load_rws_file_as_json("tests/good_formatting.JSON", False),
+                         ({"a": "foo", "b": "bar"}, None))
+
+    def test_load_and_format(self):
+        self.assertEqual(load_rws_file_as_json("tests/not_json.JSON", True),
+                         (None, "There was an error when loading tests/not_json.JSON;\nerror was:  Expecting value: line 1 column 1 (char 0)"))
+        self.assertEqual(load_rws_file_as_json("tests/bad_formatting.JSON", True),
+                         (None, 'Formatting for tests/bad_formatting.JSON is incorrect;\nerror was:\n-   "a": "foo", \n?              -\n+   "a": "foo",\n-     "b": "bar"\n? --\n+   "b": "bar"\n-   '))
```
cfredric, line 29: IMO it'd be nice if we could test this without having to write test-only files for it, i.e. if we could just pass in the string contents directly. That makes the tests faster (since they don't have to spin a disk) and easier to write (since they don't require new files in the repo).

The downside is that you have to split the file-reading part out of the function that verifies the content, but IMO that's not actually a bad thing. E.g.:

```python
    try:
        raw_content = Path(input_file).read_text()
    except Exception as inst:
        # If the file cannot be loaded, we will not run any other checks
        print(f"There was an error when loading {input_file};" 
              f"\nerror was:  {inst}")
        return

    (rws_sites, error) = parse_content_as_json(raw_content, with_format)
    if rws_sites == None or error != None:
        print(error)
        return
```

check_sites.py
```diff
@@ -12,12 +12,44 @@
...
+        or None and the error message if there was an error.
+
+        Args:
+            input_file: string
+            with_format: bool
+        Returns:
+            Tuple[Dict|None, string|None]
+    """
+    with open(input_file) as f:
+        loaded_file = f.read()
```
cfredric, line 35: IMO pathlib is nicer here.
```suggestion
    loaded_file = Path(input_file).read_text()
```

check_sites.py
```diff
@@ -50,28 +82,25 @@ def main():
...
+    opts, _ = getopt.getopt(args, "i:p:", ["data_directory=", "with_diff",
+                                         "with_format", "primaries="])
     for opt, arg in opts:
         if opt == '-i':
             input_file = arg
         if opt == '--data_directory':
             input_prefix = arg
         if opt == '--with_diff':
             with_diff = True
+        if opt == '--with_format':
```
cfredric, line 95: Bikeshedding, but it feels like we could make `with_format` a bit clearer. Maybe `strict_formatting` is clearer?

cfredric (approved):
LGTM apart from bug in the flag's default value

cfredric (changes requested):
Whoops, meant to request changes

cfredric (approved):
LGTM, thanks!

miketaylr (commented):
Thanks for documenting the requirements.

PR #676, opened 2024-12-10
Fix CRON

arichiv:
Let's fix the cron in https://github.com/GoogleChrome/related-website-sets/pull/279 so that it doesn't [fail every time](https://github.com/GoogleChrome/related-website-sets/actions/workflows/rws-submissions-checks.yml?page=2&query=event%3Aschedule). We should run it on pushes to the main branch as well as on a schedule (to ensure we don't miss changes to the public suffix list outside of RWS commits).

Tested by running on all pull requests in the new action before being reverted back: [success example](https://github.com/GoogleChrome/related-website-sets/actions/runs/12260738817/job/34206082969) [failure example](https://github.com/GoogleChrome/related-website-sets/actions/runs/12260718587/job/34206012016)


aselya (approved):
LGTM thanks for fixing this!

Let's make sure to check this before the next few submission review sessions to see if there's any issues.

PR #679, opened 2024-12-11
Run Python tests on PRs

arichiv:
This also cleans up the pull request job itself a bit.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,17 +1,4 @@
-# Copyright 2022 Google LLC
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     https://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-name: RWS submission checks
+name: Pull Request Validations
 on: pull_request_target
 jobs:
   PR-Actions:


.github/workflows/pr.yml
```diff
@@ -0,0 +1,52 @@
+# Copyright 2022 Google LLC
```
miketaylr, line 1: nit: 2024?

ok - i see you're just moving this, and not added net-new code. nevermind!

miketaylr (approved):
nice, thanks

.github/workflows/pr.yml
```diff
@@ -0,0 +1,52 @@
...
+      id: read_results
+      uses: andstor/file-reader-action@v1
+      with:
+        path: "results.txt"
+    - name: Failure Comment
+      if: steps.read_results.outputs.contents != 'success'
+      run: |
+       echo "The RWS JSON was not successfully validated. Here are your results:" > message.txt
+       cat results.txt >> message.txt
+    - name: Sucess Comment
```
cfredric, line 40:
```suggestion
    - name: Success Comment
```

.github/workflows/rws-submissions-checks.yml
```diff
@@ -1,63 +0,0 @@
...
-jobs:
-  PR-Actions:
-    runs-on: ubuntu-latest
-    outputs:
-      output1: ${{ steps.read_results.outputs.contents }}
-    steps:
-      - name: Get the PR version of the files
-        uses: actions/checkout@v2
-        with:
-          ref: "refs/pull/${{ github.event.number }}/merge"
```
cfredric, line 25: So this was intentionally running the main branch's validations against the PR, so that a malicious PR can't simply delete all the validations and bypass all requirements that way. Can we still do something like that, or are you saying that's unnecessary?

(I'm assuming that `actions/checkout` checks out the PR by default, although I don't see that mentioned in the README.)

.github/workflows/rws-submissions-checks.yml
```diff
@@ -1,63 +0,0 @@
...
-          path: "pull-request"
-      - name: Get the version at main
-        uses: actions/checkout@v2
-        with:
-          path: "main"
-      - name: Get necessary libraries
-        run: pip install publicsuffix2
-      - name: Content check
-        id: check
-        run: python3 main/check_sites.py -i pull-request/related_website_sets.JSON --data_directory=main --with_diff --strict_formatting > results.txt
```
cfredric, line 35: Looks like the data_directory flag is unused now, so there might be some cleanup to do (unless we have to rework this per my other comment).

.github/workflows/rws-submissions-checks.yml
arichiv: It's unneeded because the manual reviewer should check the files modified and ensure python wasn't touched. Checking out the main branch and using it doesn't prevent this, as it would allow one to attack in two PRs (one to modify python and a follow-up to update JSON).

.github/workflows/rws-submissions-checks.yml
arichiv: I think that's a nice feature to have, but it could be removed in a future PR (I want this one to stay focused on just the action).

.github/workflows/rws-submissions-checks.yml
arichiv: AFAIK it's an anti-pattern for actions to checkout main and use it, the one exception might be checking out a published package to use it in some way.

The important protection here is the GitHub action triggering on pull_request_target so that it's the main branches action code that runs.

cfredric: Fair, I'm just trying to avoid leaning on code review for security guarantees. But I suppose at a certain point that's inevitable.

cfredric:
LGTM, go ahead and merge when ready!

PR #681, opened 2024-12-11
Alert in comment on system file changes

arichiv:
Let's update the GitHub comment with an alert if system files (non-RWS-JSON) changes are in the PR.

This also fixes an issue with the name of the PR build that prevents checks from looking as though they passed.

Here's an example of the comment:
image


cfredric (approved):
LGTM, merge when ready!

PR #688, opened 2024-12-12
Pin requirements

arichiv:
We can also remove third_party since that's unused now.

We do not pin publicsuffix2 as we trust that library and do want the latest stable version for validation. The other two libraries just need to work and aren't as trusted, so we should hold them back until there's a reason to upgrade.


arichiv:
I think go/releasing/preparing#third-party-components-included-via-package-manager allows the use of the requirements.txt file without copying the code locally. I think the point of the third_party folder is to make clear (where code is included in the project directly) what we do and don't own. In Chromium, package managers aren't used so the third_party directory is used as a target to import from. Here that isn't the case so we can omit it I think.

cfredric:
SGTM, that makes sense. Raze it!

cfredric (approved):

PR #692, opened 2024-12-12
Fix checkout target

arichiv:
We want the action to initiate from the base commit on the main repo, but want to evaluate the code on the merge commit.


cfredric (approved):
Looks like this undoes part of the changes in https://github.com/GoogleChrome/related-website-sets/pull/679, so this was a pretty new regression.

PR #696, opened 2024-12-12
Use unified diff format when reporting formatting issues

cfredric:
I think this will give more context (like line numbers) when reporting formatting differences, that should make it easier to know where the problems are.


arichiv:
SGTM, does it work as expected when tested locally?

arichiv (approved):

arichiv:
Looks like the tests need to be updated

cfredric:
Hard to say if it works or not, since that depends on whether GitHub's markdown highlighter understands it. I'll make a test PR to see what it looks like after this merges to main.

cfredric:
Confirmed, the highlighting worked fine. (It wasn't quite as pretty as I expected, but I think it's still better than not having any line numbers/context.)

limck5856 (approved):

PR #703, opened 2024-12-13
Don't list files in commit message

arichiv:
It's not that helpful and could cause issues.


djbababobo2025:
> It's not that helpful and could cause issues.

PR #749, opened 2025-01-29
Refactor code.yml Action

arichiv:
The prior code.yml action was validating no RWS sets and running tests, and it did this on PRs, main, and CRON.

We want to validate all sets on main/CRON but not block PRs.

code.yml now just runs tests on PRs and main.
json.yml was renamed json_pr.yml with no changes.
json_main.yml was added, and validates all RWS sets on main and CRON.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -19,8 +19,6 @@ on:
   push:
     branches:
     - main
-  schedule:
-  - cron: '0 14 * * 2' # 2pm UTC each Tuesday
 permissions: {}
 jobs:
   Code-Actions:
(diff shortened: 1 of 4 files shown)


Vibe120 (approved):

Vibe120 (approved):

cfredric:
This makes it possible for unrelated timeouts/failures to block a PR submission, since PRs are validating the well-known files of the full set, right? I think it'd be better to make the PR workflows *only* validate things that are related to the PR itself.

cfredric:
We're also going to have to enhance the script itself to handle the legacy well-known file location, since the older set entries don't serve from related-website-set.json.

sjledoux:
I don't think that the new workflow this is adding,`json_main.yml`, is needed/helpful. Due to old RWS members that are not currently compliant with the RWS submission requirements, `check_sites.py` will consistently fail when passed the full JSON list.

arichiv:
I added https://github.com/GoogleChrome/related-website-sets/issues/750 to resolve errors in old sets. I think it's worth checking them so that we know if they drift out of compliance for some reason, but exemptions may have to be added for some legacy entries.

samanmokhtarzadeh (approved):

samanmokhtarzadeh (approved):

Alexandrag88 (approved):

Alishop19942663:
> The prior code.yml action was validating no RWS sets and running tests, and it did this on PRs, main, and CRON.
> 
> We want to validate all sets on main/CRON but not block PRs.
> 
> code.yml now just runs tests on PRs and main. json.yml was renamed json_pr.yml with no changes. json_main.yml was added, and validates all RWS sets on main and CRON.

PR #751, opened 2025-01-30
Improve "code" Action

arichiv:
Let's check all of the JSON but not force errors if it doesn't pass. Test failures still cause job failure.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -31,18 +31,8 @@ jobs:
       with:
         check-latest: true
     - uses: psf/black@stable
-    - name: Run Tests and Validate JSON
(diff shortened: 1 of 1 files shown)


sjledoux (approved):
LGTM as is, but might be nice if the `rws_test.py` and the `check_sites.py` calls had their own steps in the workflow, that way it would be easier for a reader to distinguish their respective outputs.

Mamun07gth:
[](url)SYMBOL_VERSION SUNW_1.22.5 {	# s10u8 - SunOS 5.10 (Solaris 10) patch additions

Mamun07gth:
SYMBOL_VERSION SUNW_1.22.5 {	# s10u8 - SunOS 5.10 (Solaris 10) patch additions

PR #752, opened 2025-02-04
Allow other errors to be listed when there is a formatting error

sjledoux:
Currently, if `parse_rws_json` is called on a JSON file that is readable but not properly formatted, we only print the formatting errors and don't run any of the other technical checks. This change makes it so that as long JSON file is readable, we still attempt to run all other technical checks so that submitters can see if their submission is invalid in other ways aside from their formatting.

### DIFF
--- a/check_sites.py
+++ b/check_sites.py
@@ -108,7 +108,7 @@ def main():
         (rws_sites, error) = parse_rws_json(rws_json_string, strict_formatting)
     except Exception as inst:
         # If the file cannot be loaded, we will not run any other checks
-        print (f"There was an error when parsing the JSON;\nerror was:  {inst}")
+        print(f"There was an error when parsing the JSON;\nerror was:  {inst}")
         return
     if error is not None:
         error_texts.append(error)
--- a/check_sites.py
+++ b/check_sites.py
@@ -72,6 +72,7 @@ def find_diff_sets(old_sets, new_sets):
     }
     return diff_sets, subtracted_sets
 
+
 def run_nonbreaking_checks(rws_checker, rws_json_string, strict_formatting, check_sets):
     """Runs all checks from check_sites and RWSCheck whose exceptions should
     not cause the program to immediately exit.
@@ -148,7 +149,6 @@ def main():
         print(f"There was an error when parsing the JSON;\nerror was:  {inst}")
         return
 
-
     # Load the etlds from the public suffix list
     etlds = PublicSuffixList(psl_file="effective_tld_names.dat")
     # Get all the ICANN domains
@@ -166,7 +166,7 @@ def main():
         # If the schema is invalid, we will not run any other checks
         print(inst)
         return
-    
+
     error_texts = []
     check_sets = {}
     subtracted_sets = {}
--- a/check_sites.py
+++ b/check_sites.py
@@ -72,8 +72,7 @@ def find_diff_sets(old_sets, new_sets):
     }
     return diff_sets, subtracted_sets
 
-def run_nonbreaking_checks(rws_checker, rws_json_string,
-                               strict_formatting, check_sets):
+def run_nonbreaking_checks(rws_checker, rws_json_string, strict_formatting, check_sets):
     """Runs all checks from check_sites and RWSCheck whose exceptions should
     not cause the program to immediately exit.
 
@@ -90,10 +89,11 @@ def run_nonbreaking_checks(rws_checker, rws_json_string,
         [String]
     """
     error_texts = []
-    if strict_formatting and (format_diff := find_format_diff(rws_json_string,
-                                                              rws_checker.rws_sites)):
-            error_texts.append(format_diff)
-    
+    if strict_formatting and (
+        format_diff := find_format_diff(rws_json_string, rws_checker.rws_sites)
+    ):
+        error_texts.append(format_diff)
+
     try:
         rws_checker.check_exclusivity(rws_checker.load_sets())
     except Exception as inst:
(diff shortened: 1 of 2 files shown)


check_sites.py
```diff
@@ -25,20 +25,18 @@
 def parse_rws_json(rws_json_string, strict_formatting):
     """Attempts to parse `rws_json_string` as JSON and validate formatting if `strict_formatting` is true.
 
-    Returns a tuple of the JSON dict and None if there were no errors,
-    or None and the error message if there was an error.
+    Returns a tuple of the JSON dict and None if there were no formatting
+    errors, or None and the error message if there was an error.  If the file
+    is not proper json, `json.loads()` will throw a JSONDecodeError.
```
cfredric, line 30: Nit: Python docs typically say "raise" instead of "throw", since that's what the Python keyword is.

```suggestion
    is not proper json, `json.loads()` will raise a JSONDecodeError.
```

check_sites.py
```diff
@@ -25,20 +25,18 @@
...
         strict_formatting: bool
     Returns:
         Tuple[Dict|None, string|None]
     """
-    try:
-        rws_sites = json.loads(rws_json_string)
-    except Exception as inst:
-        # If the file cannot be loaded, we will not run any other checks
-        return (None, f"There was an error when parsing the JSON;\nerror was:  {inst}")
+    # If this fails, it needs to be caught in the caller.
```
cfredric, line 38: I don't think we need this comment, since we've added an equivalent notice in the docstring.

check_sites.py
```diff
@@ -25,20 +25,18 @@
...
+    rws_sites = json.loads(rws_json_string)
     # Notify of any formatting errors in the JSON
     if strict_formatting:
```
cfredric, line 41: I think it'd be cleaner to split this function apart; it's really doing two things right now, parsing the JSON and validating the format. That way we can handle the different errors differently (without needing to care whether the error was raised vs returned):
* an error returned by `parse_rws_json` causes us to bail out completely, since we can't get the semantic info we need to proceed;
* an error returned by `check_formatting` (e.g.) is nonfatal and we can continue validating the semantics.

The approach in this PR works, but only because it makes a distinction between errors that are raised vs errors that are returned. IMO that's somewhat subtle and brittle, and it makes the code a bit inconsistent in the ways it indicates something went wrong.

WDYT?

check_sites.py
```diff
@@ -120,14 +130,6 @@ def main():
...
 
     rws_checker = RwsCheck(rws_sites, etlds, icanns)
-    error_texts = []
-
-    try:
-        rws_checker.validate_schema("SCHEMA.json")
-    except Exception as inst:
-        # If the schema is invalid, we will not run any other checks
-        print(inst)
-        return
```
cfredric, line 130: Does this code have to move, or can we leave it where it is (just for the sake of minimizing the diff in this PR)?

tests/rws_tests.py
cfredric, line 1: Please also add a test to verify that both formatting and semantic errors can be reported at the same time. (I.e., a regression test for this bug.)

arichiv:
The change looks good, but I'm realizing we lack an end-to-end test on main which means we don't really have a test for what happens when formatting and schema errors both exist. Could you split the 'post argument parsing' bits of main into a new function that could have unit tests added for just that purpose?

As a bonus, we might want to test what happens when the script is invoked 'with_diff' and there's an error in the file from trunk, we should probably skip the non-trunk file entirely and just note an admin should take a look.

check_sites.py
sjledoux: I think that makes sense. However, the parsing portion of this function is just the `json.loads()` call, right? Thoughts on just pulling that line into `main` within a `try ... except` block, and turning this function into `format_rws_json()`?

cfredric: Yeah, fair point :)

sjledoux: Pulled the parsing into main and renamed the function.

check_sites.py
sjledoux: Returned this to its previous location.

tests/rws_tests.py
sjledoux: This is now tested in the `testTechnicalAndFormatErrors`  unit test.

sjledoux:
@arichiv 
> The change looks good, but I'm realizing we lack an end-to-end test on main which means we don't really have a test for what happens when formatting and schema errors both exist. Could you split the 'post argument parsing' bits of main into a new function that could have unit tests added for just that purpose?

Created  the function `run_nonbreaking_checks` in `check_sites.py`, which runs the formatting check along with all of the RwsCheck technical checks that print out errors together. Also added the `TestRunNonbreakingChecks` class in `tests/rws_tests.py` for unit testing the function.

> As a bonus, we might want to test what happens when the script is invoked 'with_diff' and there's an error in the file from trunk, we should probably skip the non-trunk file entirely and just note an admin should take a look.

The current behavior isn't affected by `with_diff` - the formatting check always runs on the file from the pull request. I think it makes the most sense to keep the behavior, as any formatting issues in trunk should be caught by the daily automation anyway.

tests/rws_tests.py
```diff
@@ -1613,5 +1590,198 @@ def test_unexpteced_cctTLD(self, mock_open_and_load_json):
...
+@@ -1,5 +1,5 @@
+ {
+-  "sets":[
++  "sets": [
+     {
+       "primary": "https://primary4.com",
+       "associatedSites": [
+
+```""",
+                "There is no provided rationale for https://associated3.com",
```
arichiv, line 1759: should we have more of a preamble before the non-formatting errors? Maybe we could make each section have its own ###title to set them apart or something

sjledoux: We could add a preamble, but I think the formatting of that would take place in `main()` where we print the error messages out, so it wouldn't be visible to these unit tests.

Would be happy to tackle that in a follow-up PR.

arichiv (approved):

PR #782, opened 2025-03-05
Update .python-version

arichiv:
demo change to sign CLA


.python-version
```diff
@@ -1 +1 @@
-3.13
+9000
```
miketaylr, line 1: Welcome back from the future, Ari.

PR #789, opened 2025-03-17
Pin workflow deps to specific commits

cfredric:


arichiv:
Could we leave the actions/* ones alone? These are first party GitHub manages, the others make sense though

arichiv (changes requested):
See above

.github/workflows/json.yml
```diff
@@ -19,14 +19,14 @@ jobs:
   JSON-Actions:
     runs-on: ubuntu-latest
```
arichiv, line 20: ubuntu-24.04

.github/workflows/code.yml
```diff
@@ -26,11 +26,11 @@ jobs:
   Code-Actions:
     runs-on: ubuntu-latest
```
arichiv, line 27: ubuntu-24.04

arichiv (changes requested):

miketaylr (approved):

cfredric:
Reverted the `actions/` versions

arichiv (approved):

PR #792, opened 2025-03-17
Pin devcontainer version of black

arichiv:


.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,7 @@
   "name": "Default",
   "image": "mcr.microsoft.com/devcontainers/python:3.13",
   "features": {
-    "ghcr.io/devcontainers-extra/features/black:latest": {}
+    "ghcr.io/devcontainers-extra/features/black:edca8561d97a2e5a40ebeed64bd6660a94a1df38725a1d3f6751c6f3b60e0e30": {}
```
cfredric, line 5:
```suggestion
    "ghcr.io/devcontainers-extra/features/black:8a737e727ac5ab2f1d4cf5876720ed276dc8dc4b": {}
```

Can we use the same commit as is used in code.yml?

cfredric (changes requested):
(See my other comment)

.devcontainer/devcontainer.json
arichiv: No, they are different things. This is an docker image while that's a GitHub action

Jazzyjgirl1976 (approved):

PR #901, opened 2025-07-09
Purina related sites Update

Snagglecat:
This update adds the Purina.com and related sites to the main json file


Snagglecat:
We are going to remove the sub domains and try again

PR #915, opened 2025-07-18
undertale.wiki and deltarune.wiki

KockaAdmiralac:
I ran `check_sites.py -p https://undertale.wiki` already, but I'm not sure if that will be enough for CI to pass.


moynulislammonir252-cyber:
Hjl

PR #920, opened 2025-07-22
Add timeout to requested JSON

### LINKED ISSUE #857: Stop blocking on failing fetches
We could keep a list of failing ones or use asyncio with better timeouts to resolve the delay here.

arichiv:
We're using requests.get(...) with a 10 second timeout elsewhere, but at this one spot we were using urllib with no timeout. Let's fix that for consistency and resolve this hanging code.

Closes #857

### DIFF
--- a/RwsCheck.py
+++ b/RwsCheck.py
@@ -15,8 +15,6 @@ import json
 import requests
 
 from jsonschema import validate
-from urllib.request import urlopen
-from urllib.request import Request
 from publicsuffixlist import PublicSuffixList
 
 from RwsSet import RwsSet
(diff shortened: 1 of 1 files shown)


RwsCheck.py
```diff
@@ -290,23 +288,21 @@ def find_invalid_eTLD_Plus1(self, check_sets):
...
-        with urlopen(req) as json_file:
-            return json.load(json_file)
+        return requests.get(url, timeout=10, headers={"User-Agent": "Chrome"}).json()
 
     def check_list_sites(self, primary, site_list):
         """Checks that sites in a given list have the correct primary on their
         well-known page
 
-        Calls urlopen on a given list of sites, reads their json, and adds any
+        Calls requests.get on a given list of sites, reads their json, and adds any
```
cfredric, line 305: Nit: let's not document the implementation details of a method that this relies on, since (as this PR shows) that's liable to change without changing the core functionality of this method.

```suggestion
        Calls self.open_and_load_json on a given list of sites, reads their json, and adds any
```

PR #929, opened 2025-08-04
RWS entry for Major League Baseball

wcolodette:


arichiv:
Everything seems to be in order, once you sign the CLA I'll commit this change

related_website_sets.JSON
```diff
@@ -1066,6 +1066,16 @@
...
       }
+    },
+    {
+      "contact": "related-website-sets@mlb.com",
+      "primary": "https://mlb.com",
+      "serviceSites": [
+        "https://mlbstatic.com"
+      ],
+      "rationaleBySite": {
+        "https://mlbstatic.com": "Cookieless domain for serving static assets and images for MLB"
```
arichiv, line 1077: Why use RWS if this domain doesn't use cookies?

wcolodette: The domain had been included on the MDL, assuming because we serve our analytics and advertising scripts from the domain. We added this RWS to get off of the [MDL](https://github.com/GoogleChrome/ip-protection/blob/main/Masked-Domain-List.md) list. We added this RWS to reduce the risk of future versions of incognito Chrome from blocking [MLB.com](http://mlb.com/)'s [MLBStatic.com](http://mlbstatic.com/) dependencies

miketaylr:
@wcolodette FYI, you will need to sign the CLA before we can merge this.

